PR #106, opened 2014-05-24
added next/previous weekday functionality

agapy:
I have added two functions which return e.g. next Monday, second next Monday, previous Friday etc.


andrewelkins:
I agree with @kmonsoor his understanding is exactly how I took it.

bintoro:
Well, Arrow is already using "weekday" to mean "day of the week", so the proposed naming here is just consistent with that. But perhaps `find_weekday` would be more appropriate, and it would allow a single function to supplant the proposed two. The `weeks` argument could then be either positive or negative. Also, the `weekday` argument should be an integer, not a non-localized string.

PR #128, opened 2014-08-05
Another way to parse/format using standard timetamp formats (ISO 8601, RFC 2822)

ktossell:
This should probably use more of arrow's existing parsing and formatting code, and it would be nice if the instances of arrow.formats.Format would contain the parsing/formatting code; then users could define their own Formats.

RFC2822 timestamps need to use English day and month names.

### DIFF
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -11,6 +11,7 @@ from __future__ import absolute_import
 from arrow.arrow import Arrow
 from arrow import parser
 from arrow.util import isstr
+from arrow.formats import Format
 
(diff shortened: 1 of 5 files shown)


systemcatch:
hey @ktossell I think this functionality is already present in arrow.

```shell
>>> test = arrow.get()
>>> test.format("ddd, D MMM YYYY HH:mm:ss Z")
'Sun, 4 Aug 2019 20:13:56 +0000'
```
Feel free to reopen if you feel otherwise.

PR #151, opened 2014-08-13
Use correct Dutch wording for "hours"

wbolster:
In Dutch, it is common to use "3 uur" (singular noun) when referring to 3 hours, while we use plural forms for other time frames (days, minutes, seconds, and so on).


wbolster:
Ping, can someone merge this in please?

Issue #154, opened 2014-08-16
.to() is incompatible with pytz.timezone


ograycode:
It looks like at least ``to()`` is incompatible with pytz.timezone. This is especially concerning because pytz is so widely used, and integrated into Django. Because of this, arrow is incompatible with [djanog.utils.timezone.get_default_timezone](https://sourcegraph.com/github.com/django/django/.PipPackage/Django/.def/django/utils/timezone/get_default_timezone) 

It is currently PDT time (offset of -7 from UTC).
```
In [3]: import arrow

In [4]: arrow.now()
Out[4]: <Arrow [2014-08-16T14:04:24.883067-07:00]>

In [5]: str(arrow.now().datetime)
Out[5]: '2014-08-16 14:04:37.988805-07:00'

In [6]: str(arrow.now())
Out[6]: '2014-08-16T14:04:47.148559-07:00'

In [7]: import pytz

In [10]: str(arrow.utcnow().to(pytz.timezone("America/Los_Angeles")).datetime)
Out[10]: '2014-08-16 14:10:52.073869-08:00' # Should be -7

In [11]: str(arrow.utcnow().to(pytz.timezone("America/Los_Angeles")))
Out[11]: '2014-08-16T14:11:05.067117-08:00' # Should be -7

In [12]: str(arrow.utcnow().to("America/Los_Angeles").datetime)
Out[12]: '2014-08-16 14:11:29.675558-07:00'

In [13]: str(arrow.utcnow().to("America/Los_Angeles"))
Out[13]: '2014-08-16T14:12:55.047974-07:00'
```

This was discovered using pytz==2014.2 and arrow==0.4.4.

dekoza:
Any update on this?

dekoza:
My results are even better:
str(arrow.utcnow().to(pytz.timezone("America/Los_Angeles")).datetime)
# Returns '2015-05-29 05:03:07.061851-07:53' # wait, what?!

pytz = 2015.2
arrow = 0.5.4

andrewelkins:
I'd welcome a PR to address this @dekoza @ograycode @Exirel

ograycode:
Aside from the merge conflicts, #155 resolves this but adds a dependency on pytz.

Kudo:
Thanks @andrewelkins's merging for #254.
If #254 did fix this issue, maybe we can further close this.
Thanks.

ColtonProvias:
[...]

PR #155, opened 2014-08-18
Added pytz.timezone support.

ograycode:
This is for issue [#154](https://github.com/crsmithdev/arrow/issues/154) in which arrow did not support pytz.timezone being used.

Let me know if anything else needs to be done so that this can be merged.

I added tests, but always like to verify, just in case:

```
>>> import pytz
>>> import arrow
>>> str(arrow.utcnow().to(pytz.timezone("America/Los_Angeles")).datetime)
'2014-08-17 21:41:00.154590-07:00'
>>> str(arrow.utcnow().to(pytz.timezone("America/Los_Angeles")))
'2014-08-17T21:41:10.141368-07:00'
>>> str(arrow.utcnow().to("America/Los_Angeles").datetime)
'2014-08-17 21:41:57.354034-07:00'
>>> str(arrow.utcnow().to("America/Los_Angeles"))
'2014-08-17T21:42:03.748428-07:00'
>>> str(arrow.util.astimezone(datetime.datetime(2014, 1, 1, tzinfo=pytz.utc), pytz.timezone("America/Los_Angeles")))
'2013-12-31 16:00:00-08:00'
>>> str(arrow.util.astimezone(datetime.datetime(2014, 7, 1, tzinfo=pytz.utc), pytz.timezone("America/Los_Angeles")))
'2014-06-30 17:00:00-07:00'
```

### DIFF
--- a/arrow/util.py
+++ b/arrow/util.py
@@ -37,4 +37,22 @@ except NameError: #pragma: no cover
         return isinstance(s, str)
 
 
-__all__ = ['total_seconds', 'isstr']
+def astimezone(date, tz):
+    ''' Returns a ``datetime`` object, adjusted to the specified tzinfo.
+
+        :param date: a ``datetime`` object.
+        :param tz: a ``tzinfo`` object.
+
+    '''
+
+    date = date.astimezone(tz)
+
+    if hasattr(tz, 'normalize'):
+        # A pytz function
+        # See http://pytz.sourceforge.net/#introduction
+        date = tz.normalize(date)
+
+    return date
+
+
+__all__ = ['total_seconds', 'isstr', 'astimezone']
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -68,7 +68,7 @@ class Arrow(object):
         '''
 
         utc = datetime.utcnow().replace(tzinfo=dateutil_tz.tzutc())
-        dt = utc.astimezone(dateutil_tz.tzlocal() if tzinfo is None else tzinfo)
+        dt = util.astimezone(utc, dateutil_tz.tzlocal() if tzinfo is None else tzinfo)
 
         return cls(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
             dt.microsecond, dt.tzinfo)
@@ -455,10 +455,10 @@ class Arrow(object):
         if not isinstance(tz, tzinfo):
             tz = parser.TzinfoParser.parse(tz)
 
-        dt = self._datetime.astimezone(tz)
+        dt = util.astimezone(self._datetime, tz)
 
         return self.__class__(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
-            dt.microsecond, tz)
+            dt.microsecond, dt.tzinfo)
 
     def span(self, frame):
         ''' Returns two new :class:`Arrow <arrow.arrow.Arrow>` objects, representing the timespan
@@ -577,7 +577,7 @@ class Arrow(object):
 
         if other is None:
             utc = datetime.utcnow().replace(tzinfo=dateutil_tz.tzutc())
-            dt = utc.astimezone(self._datetime.tzinfo)
+            dt = util.astimezone(utc, self._datetime.tzinfo)
 
         elif isinstance(other, Arrow):
             dt = other._datetime
@@ -586,7 +586,7 @@ class Arrow(object):
             if other.tzinfo is None:
                 dt = other.replace(tzinfo=self._datetime.tzinfo)
             else:
-                dt = other.astimezone(self._datetime.tzinfo)
+                dt = util.astimezone(other, self._datetime.tzinfo)
 
         else:
             raise TypeError()
@@ -732,7 +732,7 @@ class Arrow(object):
 
         '''
 
-        return self._datetime.astimezone(tz)
+        return util.astimezone(self._datetime, tz)
 
     def utcoffset(self):
         ''' Returns a ``timedelta`` object representing the whole number of minutes difference from UTC time. '''
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ nose==1.3.0
 nose-cov==1.6
 chai==0.4.0
 sphinx==1.2b1
+pytz==2014.4
--- a/tests/util_tests.py
+++ b/tests/util_tests.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 
 from chai import Chai
-from datetime import timedelta
+from datetime import timedelta, datetime
 import sys
+import pytz
 
 from arrow import util
 
@@ -26,3 +27,19 @@ class UtilTests(Chai):
 
             assertEqual(util._total_seconds_27(td), 30)
 
+    def test_astimezone(self):
+        tz = pytz.timezone("America/Los_Angeles")
+
+        # Test outside of daylight savings time
+        dt = datetime(2014, 1, 1, tzinfo=pytz.utc)
+        dt = util.astimezone(dt, tz)
+        
+        assertFalse(dt.dst())
+        assertEqual(dt.tzinfo._tzinfos, tz._tzinfos)
+
+        # Test inside of daylight savings time
+        dt = datetime(2014, 7, 1, tzinfo=pytz.utc)
+        dt = util.astimezone(dt, tz)
+
+        assertTrue(dt.dst())
+        assertEqual(dt.tzinfo._tzinfos, tz._tzinfos)


chebee7i:
#72, #154 

It seems like arrow should consider using `pytz` only, possibly also using https://github.com/regebro/tzlocal/blob/master/tzlocal/windows_tz.py for conversion to/from Windows timezone names.

For a nice explanation of the problems with dateutil, see: http://www.assert.cc/2014/05/25/which-python-time-zone-library.html

Also relevant: http://stackoverflow.com/tags/timezone/info

Syeberman:
This issue might be obsolete, as it sounds like `dateutil` can automatically update from the IANA database: https://dateutil.readthedocs.io/en/stable/#building-and-releasing

(Note the top Google hit for `dateutil` is quite out of date: https://labix.org/python-dateutil)

Further, PyPI has a version from April of this year, and appears to be updated a few times a year: https://pypi.python.org/pypi/python-dateutil/2.5.3

Perhaps this PR can be closed?

chebee7i:
How does that make this issue obsolete?  There are very good arguments for why `pytz` should be preferred over `dateutil` and the fact that `dateutil` can autoupdate from IANA does not address those issues.

andrewelkins:
I'll review this one tonight.

pganssle:
FWIW, I do not think there are any longer particularly strong arguments for the use of `pytz` over `dateutil`, as of dateutil version 2.6.0 (which came out after [the linked blog post above](http://www.assert.cc/2014/05/25/which-python-time-zone-library.html). The issue with ambiguous times has been fixed in a way that is compatible with PEP 495 (the forthcoming official Python fix for this issue), and `dateutil` time zones are designed using the intended `tzinfo` interface, so you can safely do things like:

```python
from dateutil import tz
from datetime import datetime
datetime(2016, 11, 2, 1, 30, tzinfo=tz.gettz('America/New_York')
```

and it will work as intended. You can also then do limited arithmetical operations on those. In general, as of Python 2.6.0, I would think that `dateutil` time zones should be preferred over `pytz` in most cases (there are some cases where attaching what is effectively a static zone would be preferable, but there are often other solutions to this that would work better anyway). As of Python 3.6, the [datetime documentation](https://docs.python.org/3.6/library/datetime.html#tzinfo-objects) has switched from recommending `pytz` to `dateutil.tz` in the "See also" box.

This is not to say that `pytz` zones should *not* be supported, just that many of the objections to dateutil zones has been alleviated recently, and I would not recommend switching to `pytz` as the default.

As for the question of IANA zones, both `pytz` and `dateutil` default to using the system zoneinfo files. The versions of the IANA database that each ships is only used as a fallback (which comes up mostly on Windows since Windows doesn't ship with its own zonefile). `dateutil` will not auto-update from IANA and unfortunately the interface is still a bit rough on how to get the updated version - you basically clone the repo, run the script and re-build it. For now we'll try and cut releases whenever there is an IANA release until an easier ("just run the script to update") process is available.

arrow/util.py
```diff
@@ -37,4 +37,22 @@ def isstr(s):
...
+        :param tz: a ``tzinfo`` object.
+
+    '''
+
+    date = date.astimezone(tz)
+
+    if hasattr(tz, 'normalize'):
+        # A pytz function
+        # See http://pytz.sourceforge.net/#introduction
+        date = tz.normalize(date)
```
pganssle: This is not necessary - `astimezone(tz)` should always produce an unambiguous date, even with `pytz`. You only need to use `normalize` if you have done an arithmetical operation, in case this has resulted in the date crossing a boundary.

ograycode:
Hey everyone, this popped up on my notification list today. I don't have any plans to rebase or otherwise fix the merge conflicts. I never used arrow because of the issue that this pull request attempted to solve in 2014. Instead, the company I work for developed a set of wrappers and utility functions around `datetime` to use in the django projects we do. You can view it at https://github.com/ZeroCater/zc_common/blob/master/zc_common/timezone.py.

pganssle:
I think the *only* change that needs to happen for `pytz` to work is [this one](https://github.com/crsmithdev/arrow/pull/155/files#diff-5d8eb9e97f8b648e4dae883cdce7b959R461), FYI. The `normalize` function is not necessary if you're already calling `astimezone`, and all the other changes revolve around bringing in this `astimezone` function that calls `normalize`.

andrewelkins:
Thanks @ograycode and @pganssle  for chiming in. I'll check out that part specified by pganssle and merge.

chebee7i:
@pganssle that is fantastic news! Just to clarify, in your comment you said dateutil timezones should be preferred over pytz ones from Python 2.6.0 and higher, but I think you meant dateutil 2.6.0, which was just released on 2016-11-08. And as to the reason why they should be preferred, is that because of fold support? Or are there additional reasons as well now?

pganssle:
@chebee7i Yes, my bad. I've been getting myself confused because `dateutil 2.6.0` adds PEP-495 support while `Python 3.6.0` adds PEP-495.

As for the reasoning, there's nothing *wrong* with `pytz` per se, if you're using it correctly, but `pytz` really only provides static time zones plus some tooling around how to choose them correctly. `dateutil` time zones implement the `tzinfo` API properly such that they do not need an equivalent of `localize` or `normalize`. Consider this:

```python
import pytz
from datetime import datetime, timedelta

NYC = pytz.timezone('America/New_York')
datetime(2016, 11, 11, 12, tzinfo=NYC)
# Note how constructing a datetime using the constructor gives a very wrong answer:
# datetime.datetime(2016, 11, 11, 12, 0, tzinfo=<DstTzInfo 'America/New_York' LMT-1 day, 19:04:00 STD>)

dt = NYC.localize(datetime(2016, 5, 11, 12))
# This is better:
# datetime.datetime(2016, 5, 11, 12, 0, tzinfo=<DstTzInfo 'America/New_York' EDT-1 day, 20:00:00 DST>)

dt.replace(month=11)
# Should be EST, but it's EDT
# datetime.datetime(2016, 11, 11, 12, 0, tzinfo=<DstTzInfo 'America/New_York' EDT-1 day, 20:00:00 DST>)

dt + timedelta(days=184)
# Again, should be EST, but it's EDT
# datetime.datetime(2016, 11, 11, 12, 0, tzinfo=<DstTzInfo 'America/New_York' EDT-1 day, 20:00:00 DST>)

NYC.normalize(dt.replace(month=11))
# There we go, fixed it
# datetime.datetime(2016, 11, 11, 11, 0, tzinfo=<DstTzInfo 'America/New_York' EST-1 day, 19:00:00 STD>)
```

As you can see, there's a way to do things, but the obvious way to do things isn't the *right* way to do them, usually, because `pytz`'s time zone API is designed with a different philosophy than Python's time zones. Compare to `dateutil`:

```python
from datetime import datetime, timedelta
from dateutil import tz

NYC = tz.gettz('America/New_York')
dt = datetime(2016, 11, 11, 12, tzinfo=NYC)
print(str(dt) + ' ' + dt.tzname())
# Works correctly when passed to the constructor
# 2016-11-11 12:00:00-05:00 EST

dt2 = dt - timedelta(days=184)
print(str(dt2) + ' ' + dt2.tzname())
# Time zone addition works as well
# 2016-05-11 12:00:00-04:00 EDT
```

So I'd say that `dateutil`'s time zones are a more "pythonic" time zone interface, but of course `pytz` has had a *feature-complete* IANA time zone interface for much longer, so a lot of people have gotten used to doing it the `pytz` way. I am also fairly certain that the way `pytz` does things is less resource-intensive in some cases, because the pythonic interface involves calculating the UTC offset as a function of date every time you want it. `pytz` sacrifices some flexibility, but they calculate the exact UTC offset / tzname / DST status only when `localize` is called, and after that the value is cached when you try and pull it up.

That said, it's not a particularly expensive calculation, and some `dateutil` zones will actually cache the lookup automatically, behind the scenes (e.g. `dateutil.tz.tzical` zones), and eventually that kind of caching could be (and likely will be at some point) implemented in `tzfile`. I didn't implement any caching when I updated `tzfile` mostly because `tzfile` *used* to use a much less efficient algorithm for looking up the UTC offset even *without* caching (it was using a linear search on a sorted list rather than using `bisect`) and no one ever complained about performance, so I figured there was no urgent need to optimize beyond the immediate speedup provided by switching to `bisect`.

treeandbrick:
Helpful, thanks. I'm familiar with the particular quirks you mentioned---recommended practice for `pytz` is to use localize and normalize (when appropriate). So when adopting that practice, you don't hit those quirks. But yes, in agreement wrt the more pythonic interface of dateutil.

Aside: @abalkin can you comment on the reasons for changing the recommendation from `pytz` to `dateutil`? I didn't see any discussion of it on https://bugs.python.org/issue27595 or in the commit itself. Mostly just curious. My experience is that `pytz` has been the de-facto standard timezone library for some time now, even though both libraries are very popular (and people often use dateutil in conjunction with pytz). With the release of dateutil 2.6.0 and the recommendation flip, it sounds like this may start to change. Definitely a good reason for `arrow` to support both.

chebee7i:
Helpful, thanks. I'm familiar with the particular quirks you mentioned---recommended practice for `pytz` is to use localize and normalize (when appropriate). So when adopting that practice, you don't hit those quirks. But yes, in agreement wrt the more pythonic interface of dateutil.

Aside: @abalkin can you comment on the reasons for changing the recommendation from `pytz` to `dateutil`? I didn't see any discussion of it on https://bugs.python.org/issue27595 or in the commit itself. Mostly just curious. My experience is that `pytz` has been the de-facto standard timezone library for some time now, even though both libraries are very popular (and people often use dateutil in conjunction with pytz). With the release of dateutil 2.6.0 and the recommendation flip, it sounds like this may start to change. Definitely a good reason for `arrow` to support both.

abalkin:
> can you comment on the reasons for changing the recommendation from pytz to dateutil?

You can find relevant discussions in the datetime-sig archives.  In short, pytz violates the datetime.tzinfo design where timezone instance is supposed to be shared between datetime objects.  This was done in order to solve the problem of ambiguous local times which is now solved by PEP 495.  The pytz approach, while ingenious, was considered to be a hack by the datetime module developers (Tim Peters/@tim-one and myself.)  Meanwhile dateutil has caught up with pytz in terms of IANA tzdata support and gained PEP 495 features.  That's why we can no longer recommend pytz in documentation.

Syeberman:
It's time to close this PR, but thanks everyone for their input!

PR #165, opened 2014-09-14
Add suport for "en_gb" locale

DTVD:
Hi! 
Firstly thanks for making this amazing useful lib.

I used `arrow` for my project [Rainbow Stream](https://github.com/DTVD/rainbowstream) and my users in `en_GB` locale complained that their locale is not supported so I created this PR.

Thanks again !


andrewelkins:
fixed with #156

PR #166, opened 2014-09-16
Added ArrowInterval class that conforms to the standards of the Joda interval (http://joda-time.sourceforge.net/apidocs/org/joda/time/Interval.html)

fatisar:


kdeldycke:
What about using vocabulary from [Allen's interval algebra](https://en.wikipedia.org/wiki/Allen%27s_interval_algebra) instead ?

Syeberman:
It looks like [Joda is deprecated](http://www.joda.org/joda-time/), and [its replacement](https://docs.oracle.com/javase/8/docs/api/java/time/package-summary.html) has no similar concept.

Arrow is inspired by Moment.js, which implements this as [methods on a moment](http://momentjs.com/docs/#/query/is-between/).  This might be a better direction for Arrow.

andrewelkins:
Can confirm @Syeberman thoughts. We would prefer to stick close to moment.js where possible.

Syeberman:
Thanks for the contribution @fatisar, and sorry for the delay, but this isn't the direction we want to take Arrow.

PR #187, opened 2014-12-24
Update humanize local describe for Chinese CN/TW

axiaoxin:
Update humanize local describe for Chinese CN/TW, "秒前" is not a correct Chinese grammar


andrewelkins:
No response

PR #190, opened 2015-01-06
Parse lower-cased 'h'

tamentis:
It appears the documented 'h' and 'hh' tokens are not currently working, the following Pull Request is an attempt at fixing that.

Before the fix:

```python
>>> arrow.get("5 PM", "h A")
<Arrow [0001-01-01T12:00:00+00:00]>
```

After the fix:
```python
>>> arrow.get("5 PM", "h A")
<Arrow [0001-01-01T17:00:00+00:00]>
```

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -175,7 +175,7 @@ class DateTimeParser(object):
         elif token in ['DD', 'D']:
             parts['day'] = int(value)
 
-        elif token in ['HH', 'H', 'hh', 'h']:
+        elif token.upper() in ['HH', 'H']:
             parts['hour'] = int(value)
 
         elif token in ['mm', 'm']:
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -175,7 +175,7 @@ class DateTimeParser(object):
         elif token in ['DD', 'D']:
             parts['day'] = int(value)
 
-        elif token in ['HH', 'H']:
+        elif token in ['HH', 'H', 'hh', 'h']:
             parts['hour'] = int(value)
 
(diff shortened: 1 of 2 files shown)


arrow/parser.py
```diff
@@ -175,7 +175,7 @@ def _parse_token(self, token, value, parts):
         elif token in ['DD', 'D']:
             parts['day'] = int(value)
 
-        elif token in ['HH', 'H']:
+        elif token in ['HH', 'H', 'hh', 'h']:
```
pypingou: Might be just as easy to do a ``token.upper()`` now?

tamentis: But that would break 's' and 'S'.

pypingou: On Wed, Jan 07, 2015 at 05:37:44PM -0800, Bertrand Janin wrote:
>    In arrow/parser.py:
> 
>  > @@ -175,7 +175,7 @@ def _parse_token(self, token, value, parts):
>  >          elif token in ['DD', 'D']:
>  >              parts['day'] = int(value)
>  > 
>  > -        elif token in ['HH', 'H']:
>  > +        elif token in ['HH', 'H', 'hh', 'h']:
> 
>    But that would break 's' and 'S'.

Depends if you do at the top an ``token = token.upper()`` or just adjust the
if/elif statements to be ``elif token.upper() in ['HH', 'H']``

tamentis: Updated that as per your comment, however I can't use the same logic for dd/d since they can't be parsed the same way as DD/D (ints VS names). I'll give that a shot in a separate pull request.

PR #163, opened 2014-09-11
Fix for ISO parsing and formatting for fractional seconds

sdispater:
This pull request fixes the bug referenced in #83.

Now, fractional seconds are properly handled and not considered as microseconds:
```python
arrow.get('2013-05-05T12:30:45.2+0030').datetime
datetime.datetime(2013, 5, 5, 12, 30, 45, 200000, tzinfo=tzoffset(None, 1800))

arrow.get('2013-05-05T12:30:45.002+0030').datetime
datetime.datetime(2013, 5, 5, 12, 30, 45, 2000, tzinfo=tzoffset(None, 1800))

arrow.get('2013-05-05T12:30:45.002+0030').format('YYYY-MM-DD HH:mm:ss.SSS ZZ')
'2013-05-05 12:30:45.002 +00:30'

arrow.get('2013-05-05T12:30:45.002+0030').format('YYYY-MM-DD HH:mm:ss.SSSSS ZZ')
'2013-05-05 12:30:45.00200 +00:30'

arrow.get('2013-05-05T12:30:45.002+0030').format('YYYY-MM-DD HH:mm:ss.SSSSSS ZZ')
'2013-05-05 12:30:45.002000 +00:30'
```

PR #192, opened 2015-01-22
Handle fractional seconds with padded 0's

andrewelkins:
Handles fractional seconds with padded 0's as documentation on other values within the formatter do. 

http://crsmithdev.com/arrow/#tokens


andrewelkins:
This is also addressed by pull #163 Although the changes I made match the rest of the file with regard to formatting the functions used. (Using format vs zfill) So you may want to pull both.

andrewelkins:
Closing, it'll be covered in #163

PR #200, opened 2015-02-26
Case insensitive month match

cshowe:
arrow.get() requires an exact string match for the month name and month abbreviation tokens (MMM and MMMM).  Thus Feb matches MMM but feb and FEB do not.  This is straightforward to work around in calling code, but seemed as though it might be common enough to warrent a patch.

Note that this does not work with non english locales (due to using calendar to build the regex rather than the parser locale) but as far as I can tell this isn't a regression.


andrewelkins:
Thanks for adding the tests. Once those come back positive I'll merge.

PR #236, opened 2015-07-01
import timedelta in __init__.py

Riamse:
This makes things less annoying by allowing the user to import timedelta from arrow itself instead of arrow.util or datetime.


andrewelkins:
But then people would rely on arrow to maintain this inclusion. It might be a little more hassle, but I don't think we want to do that.

Issue #174, opened 2014-10-27
.get() parse gives wrong result with strings in yyyy/mm/dd format


jejansse:
In version 0.4.2 the .get() function parsed strings in the yyyy/mm/dd format in a correct manner. In version 0.4.4, however, these strings get parsed as yyyy-01-01. For example "2014/10/27" now gets parsed as "2014/01/01", whereas in previous versions it parsed correctly. I guess this is a bug?

Here is a code example:

```python
>>> import arrow
>>> arrow.get("2014-10-27")
<Arrow [2014-10-27T00:00:00+00:00]>
>>> arrow.get("2014/10/27")
<Arrow [2014-01-01T00:00:00+00:00]>
```

PR #242, opened 2015-07-10
added general datetime formats

andrewelkins:
Addresses #174


andrewelkins:
@arvindch Can you add tests for this?

arvindch:
We can close this now, because of #271.

PR #248, opened 2015-07-22
Adding Belarusian (be) locale

Oire:


pypingou:
Isn't ``be`` for belgium ?

Oire:
@pypingou, there's no Belgian language :). The two-character code for Belarusian is `be`. Does Arrow use other codes for languages?

pypingou:
My bad, it seems Belgium is using fr_be and nl_be indeed, thanks :)

PR #249, opened 2015-07-24
Search date in strings

beenje:
Unknown tokens in the string are ignored.

In the previous implementation, each token was matched individually.
With the format 'YYYY-MM-DD', parsing the string '2000 ... 2015-12-31'
would fail because 'YYYY' taken alone matches the first 4 numbers.

Same with the format 'MMMM YYYY' and the string 'June was born in May
1980'. 'MMMM' would match 'June' instead of 'May' even if there are no
digits next to it.

This change builds a complete pattern string by replacing each token by
its pattern. 'YYYY-MM-DD' becomes '(?P<YYYY>\d{4})-(?P<MM>\d{2})-(?P<DD>\d{2})'.
The search is done with that string.
That matches '2015-12-31' in the previous example.

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -4,8 +4,6 @@ from __future__ import unicode_literals
 
 from datetime import datetime
 from dateutil import tz
-
-import calendar
 import re
 
 from arrow import locales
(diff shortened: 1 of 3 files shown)


beenje:
I just realized that "fuzzy parsing" might not be the correct term.
I should have named that "matching date in long strings".

beenje:
I changed the title and removed mentions to fuzzy.

PR #253, opened 2015-08-01
Raise TypeError when adding/subtracting non-dates

### LINKED ISSUE #241: NotImplementedError should be TypeError when adding and subtracting
Here is the default behavior in Python, which is what you would expect:

```
>>> 5 - None
TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'
```

Using `*` or `/` with an arrow object yields the default behavior:

```
>>> arrow.utcnow() * 7
TypeError: unsupported operand type(s) for *: 'Arrow' and 'int'
>>> arrow.utcnow() / 5
TypeError: unsupported operand type(s) for /: 'Arrow' and 'int'
```

But `+` and `-` throw a `NotImplementedError` with no message:

```
>>> arrow.utcnow() + None
  File "/Users/jrw/.virtualenvs/ford/lib/python2.7/site-packages/arrow-0.5.0-py2.7.egg/arrow/arrow.py", line 650, in __add__
    raise NotImplementedError()
NotImplementedError
>>> arrow.utcnow() - 5
  File "/Users/jrw/.virtualenvs/ford/lib/python2.7/site-packages/arrow-0.5.0-py2.7.egg/arrow/arrow.py", line 666, in __sub__
    raise NotImplementedError()
NotImplementedError
```

itsmeolivia:
Here's what used to happen:
```py
arrow.utcnow() + None #=> raises NotImplementedError()
```
Now: 
```py
arrow.utcnow() + None #=> raises TypeError()
```

Fixes #241.

I'm not sure this is actually what should be happening but I already wrote it when I read that TypeError should not be raised.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -669,7 +669,7 @@ class Arrow(object):
         if isinstance(other, (timedelta, relativedelta)):
             return self.fromdatetime(self._datetime + other, self._datetime.tzinfo)
 
-        raise NotImplementedError()
+        raise TypeError()
 
     def __radd__(self, other):
         return self.__add__(other)
@@ -685,7 +685,7 @@ class Arrow(object):
         elif isinstance(other, Arrow):
             return self._datetime - other._datetime
 
-        raise NotImplementedError()
+        raise TypeError()
 
     def __rsub__(self, other):
         return self.__sub__(other)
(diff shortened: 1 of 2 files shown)


andrewelkins:
https://docs.python.org/2/library/exceptions.html#exceptions.TypeError

PR #259, opened 2015-08-20
Throw TypeError when passing booleans to arrow.get

### LINKED ISSUE #238: arrow.get() converts bool types (False and True) incorrectly
``` python
In [1]: import arrow

In [2]: arrow.get(False)
Out[2]: <Arrow [1970-01-01T00:00:00+00:00]>

In [3]: arrow.get(True)
Out[3]: <Arrow [1970-01-01T00:00:01+00:00]>
```

This is because of the cast of the input to a float. Float type casts booleans to their underlying numeric representations. However i don't think true bool types should be interpreted as valid date/times.

EvanHahn:
Before:

```
>>> arrow.get(False)
<Arrow [1970-01-01T00:00:00+00:00]>

>>> arrow.get(True)
<Arrow [1970-01-01T00:00:01+00:00]>
```

After:

```
>>> arrow.get(False)
TypeError: Can't parse single argument type of '<type 'bool'>'

>>> arrow.get(True)
TypeError: Can't parse single argument type of '<type 'bool'>'
```

Closes #238.

### DIFF
--- a/arrow/util.py
+++ b/arrow/util.py
@@ -23,6 +23,8 @@ else: # pragma: no cover
     total_seconds = _total_seconds_27
 
 def is_timestamp(value):
+    if type(value) == bool:
+        return False
     try:
         float(value)
         return True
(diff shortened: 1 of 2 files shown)


EvanHahn:
Any updates on this? No problem if not.

PR #271, opened 2015-10-09
Extend parseable formats

### LINKED ISSUE #174: .get() parse gives wrong result with strings in yyyy/mm/dd format

arvindch:
Hi!

So, I've added more separators (`- / .`) for parsing to fix #174.
Additionally, I've added tests for them, as per #242.

Cheers!

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -52,6 +52,9 @@ class DateTimeParser(object):
         'S': re.compile('\d'),
     }
 
+    MARKERS = ['YYYY', 'MM', 'DD']
+    SEPARATORS = ['-', '/', '.']
+
     def __init__(self, locale='en_us'):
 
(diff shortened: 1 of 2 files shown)


andrewelkins:
Thanks @arvindch

Issue #220, opened 2015-05-13
adjusting by an offset is easily confused with adjusting a specific part of a date


mozbhearsum:
Someone pointed out to me today that having such similarly named keywords in the .replace() API has the potential for some confusion. Eg: "minute" and "minutes" where the former changes the minute the value given, while the latter adjusts by an offset. This means that very similar looking (and easy to typo) calls can do vastly different things:
arrow.now().replace(month=5)
arrow.now().replace(months=5)

This person suggested that a different method for offset adjustments might be better. Eg:
arrow.now().shift(months=5)

andrewelkins:
I like this a lot. Can you make a PR for this?

rsyring:
can this be closed now that `shift()` is a method?

systemcatch:
It can be! :hammer:

PR #276, opened 2015-10-28
Separate replace & shift function

sipp11:
According to PR Request from the issue #220, I also think it's a nice idea.

Basically ripping relativedelta part out of replace() to make a new function, shift() & refactoring all related tests.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -378,11 +378,6 @@ class Arrow(object):
         >>> arw.replace(year=2014, month=6)
         <Arrow [2014-06-11T22:27:34.787885+00:00]>
 
-        Use plural property names to shift their current value relatively:
-
-        >>> arw.replace(years=1, months=-1)
-        <Arrow [2014-04-11T22:27:34.787885+00:00]>
-
         You can also provide a timezone expression can also be replaced:
 
         >>> arw.replace(tzinfo=tz.tzlocal())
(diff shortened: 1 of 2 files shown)


andrewelkins:
Excellent, I'll review this and get it merged.

andrewelkins:
Looks solid. Let me do some testing locally before pushing out this breaking change.

sipp11:
Oops, I miss something, the docs, `docs/index.rst`. However, docstrings are good enough, I hope.

kdeldycke:
:100: : this is a huge usability gain !

pypingou:
Would there be a way to make replace call shift so that it doesn't break backward compatibility?
Maybe with a deprecation warning in addition?

sipp11:
@pypingou That shouldn't be a problem; however, I don't know if that's the best option. Let's hear what @andrewelkins says. I could do another PR if needed although I prefer the way this is.

pypingou:
@sipp11 how would it not be a problem? All the calls to ``.replace()`` will suddenly stop working, that's pretty much breaking backward compatibility :)

sipp11:
@pypingou Sorry I might reply you too short since it should have been __"That shouldn't be a problem to add deprecation warning."__ 

For the breaking change, I'm fine either way since I've fixed them all. Code is much easier to read :)

pypingou:
It's easier to read and I think the change is great but breaking backward compatibility from a release to another is bad. I do think we should announce the change with a deprecation warning for a release (ideally two), then we can clean the code and get ride of ``.replace()``.

andrewelkins:
That makes sense. Can we add the deprecation warning?

sipp11:
Sorry that I couldn't do it sooner, but if you still need it, I restore replace functionality (to shift date) and add deprecation warning if using it. Hopefully I'll get another PR done too. I use this library too much and I hate to just pip from my fork.

Syeberman:
I've resolved the merge conflicts in https://github.com/crsmithdev/arrow/issues/349

andrewelkins:
Thanks @sipp11 for the original PR and @Syeberman I'll check out the new PR and close this one.

Issue #275, opened 2015-10-22
unsupported timezones


paul-hammant:
"EDT" (any many more) are not supported in plain parsed formats, for arrow.

Perhaps link to dateutil's gettz function in the documentation? And example code like so:
corrected_arrow_date = parsed_arrow_date.replace(tzinfo=dateutil.tz.gettz("EDT"))

swistakm:
My proposition is to add new `ZZZ` token for parsing timezone names from [tz database](http://www.iana.org/time-zones) like `GMT`, `Europe/Warsaw` etc.

Rationale: even for well established date-time formats that rely on offset rather than TZ names developers often face deviations in third party data. At my work we parse a lot of RSS feeds that are supposed to use [RFC-822](http://www.w3.org/Protocols/rfc822/#z28) in `<pubDate/>` element. Often developers behind those feeds do not follow spec and use `GMT` or other picked timezone specifier. I could simply define another format for arrow if I had and token available to handle that.

Now I am forced to use this hackaround:
```python
arrow.parser.DateTimeParser._BASE_INPUT_RE_MAP['Z'] = re.compile('([\w/]+)')
```

With that I am able to parse dates like:
```
> arrow.get('22 May 2015 08:15:00 GMT', ["DD MMM YYYY HH:mm:ss Z"])
<Arrow [2015-05-22T08:15:00+00:00]>

> arrow.get('22 May 2015 08:15:00 Europe/Warsaw', ["DD MMM YYYY HH:mm:ss Z"])
<Arrow [2015-05-22T08:15:00+02:00]>
```

This works because arrow behind the scenes uses `dateutil.tz.gettz` in `TzinfoParser.parse()` method that is supposed to work with timezones from tz database.

Also additional improvements that I see:
* `DateTimeParser` could reuse regexp from `TzinfoParser` to reduce code duplication because regular expressions are also a code (domain specific but still code).
* `Z` and `ZZ` tokens parse both offset formats despite documentations says something different. I don't know when this change was introduced but now it cannot be rectified without breaking compatibility for users. There is only one solution IMO: fix documentation saying that both parse both available offset formats but give different results when formatting.

As a sidenote: @paul-hammant in my environment the call `dateutil.tz.gettz("EDT")` returns `None` and latest pytz raises `UnknownTimeZoneError` error for such timezone. So your example code do not fail but it seems that provided example `EDT` timezone is not valid.

PR #280, opened 2015-11-13
parser: add new ZZZ token to recognize/parse TZ names from the tz database

swistakm:
Solution proposed as a fix to issue #275  

* try to test various available tz naming conventions
* rename existing `test_parse_tz` to `test_parse_tz_zz` in order
  to separate ZZ tests from ZZZ
* update token table in docs
* add new regex for ZZZ token and leverage existing code that recognizes
  timezones via dateutil

This does not include improvements (refactoring) to existing code proposed also in the issue thread.

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -15,7 +15,7 @@ class ParserError(RuntimeError):
 
 class DateTimeParser(object):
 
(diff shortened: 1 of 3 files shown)


andrewelkins:
Thanks @swistakm for the PR

PR #293, opened 2016-01-04
adding arrow.quarter property and relative `replace` options

benmathes:
While I saw quarters used in `range` and `span_range`, I didn't see a way to have an instance property, e.g.

```python
a = arrow.utcnow()
print(a.quarter) # 1,2,3, or 4
```

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -408,13 +408,20 @@ class Arrow(object):
 
             if key in self._ATTRS:
                 absolute_kwargs[key] = value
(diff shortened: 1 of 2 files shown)


benmathes:
While I would have stylistically preferred a `@property`, the committed code attempted to use the current patterns. Please let me know if there's another way you'd like this.

benmathes:
Also: I did write tests for this. The build is failing due to 99% (less than required 100%) coverage from a prior commit.

Issue #291, opened 2015-12-28
ISO8601 isn't supported. Suggest stop naming it in the docs or fix it.


boxed:
Got a tip today about arrow and saw that ISO8601 is mentioned a few times in the readme. Tried two things: `arrow.get('2001-01-01')`. That gave me a datetime, but it's a date. 

Then I tried `arrow.get('2001W02')` and it gave me `<Arrow [2001-02-01T00:00:00+00:00]>`. This is wrong because "W02" means week two and not february. Pretty scary that it just silently mis-parses dates like this.

Then I tried times. `arrow.get('00:00')` which just raises a ParseError. Seems a bit weak, but at least it doesn't say 0000-01-01 :P

I have a pretty comprehensive test suite for ISO8601 on my own little project: https://github.com/boxed/iso8601/blob/master/iso8601.py (scroll to the bottom) if you want to fix this. I probably wouldn't recommend using the parsing I do in my lib though, it's super super hacky :P

boxed:
Why was this issue closed? All the issues raised above are still true and your parser is still broken.

ramonsaraiva:
@boxed docs were updated on https://github.com/crsmithdev/arrow/pull/308 according to what you've suggested on the issue title, could you please re-create an issue with a more reasonable acceptance criteria?

boxed:
Ah ok, I'll just create a new ticket with the bad parsing behavior then.

ramonsaraiva:
thank you!

PR #308, opened 2016-02-17
Adds support for tz offsets without minutes

### LINKED ISSUE #307: arrow.get() does not parse PostgreSQL timezone-aware timestamps correctly
`arrow.get()` fails to parse [PostgreSQL's timezone aware output](http://www.postgresql.org/docs/9.1/static/datatype-datetime.html#DATATYPE-DATETIME-OUTPUT) when the UTC offset is an integer number of hours.

For example:

> `arrow.get('2050-01-01 00:00:00+00')`

Raises:

> `ParserError: Could not match input to any of [u'YYYY-MM-DD HH:mm:ssZ'] on '2050-01-01 00:00:00+00'`

However, if you add ':00' to the offset, it parses successfully:

> `arrow.get('2050-01-01 00:00:00+00:00')`
> `<Arrow [2050-01-01T00:00:00+00:00]>`

According to [the wikipedia article for ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators) it's completely valid to specify the timezone offset without a minutes component (e.g. `<time>±hh`). I believe this is a bug in the Arrow library that should be corrected.

bjmc:
Fixes #307 and doc changes to address #291

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,8 @@ Features
 - Time zone-aware & UTC by default
 - Provides super-simple creation options for many common input scenarios
 - Updated .replace method with support for relative offsets, including weeks
-- Formats and parses strings, including ISO-8601-formatted strings automatically
+- Formats and parses strings automatically
+- Partial support for ISO-8601
 - Timezone conversion
 - Timestamp available as a property
 - Generates time spans, ranges, floors and ceilings in timeframes from year to microsecond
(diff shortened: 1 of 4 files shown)


bjmc:
@crsmithdev Travis failures seem to be related to coverage in locale tests, not anything in this changeset.

PR #310, opened 2016-03-04
Including BenaliLocaleTests so that the CI can stop crying

seemethere:
BengaliLocale didn't have tests so I added them.


seemethere:
Didn't notice that someone already did this here in #289

PR #322, opened 2016-04-14
Fix typo in doc

harcher81:


jaapz:
This is not a typo, it's the correct API for replace.

To replace the current hour in a datetime with a different hour, you use:
    
    date = arrow.utcnow()
    date.replace(hour=4)

To *add* or *subtract* a number of hours from the current hour, you'll use `hours`, like this:

    date = arrow.utcnow()
    date.replace(hours=-5)
    date.replace(hours=4)

PR #327, opened 2016-05-01
Cleaned up function structure in formatter.py

David-OConnor:
Removed the unnecessary DateTimeFormatter class; restructured formatter.py to be easy to understand.


andrewelkins:
Can you fix the unit tests?

David-OConnor:
Good call - Done.

PR #334, opened 2016-05-18
Format ParserError's raise message

youngminz:


youngminz:
Ping for if you unread and skipped this issue. :)

andrewelkins:
Good catch. That's what we use elsewhere, good to keep the style the same.

PR #342, opened 2016-06-28
Documentation: clarify timezone replacement vs conversion

Syeberman:
I'm often confused about which methods perform a timezone conversion (possibly modifying the hour and other attributes) and which simply replace the timezone (leaving all other attributes the same).  So I've updated the documentation to clarify this.

I've also made other improvements to the documentation, which I'll describe in-line.

### DIFF
--- a/arrow/api.py
+++ b/arrow/api.py
@@ -13,18 +13,17 @@ from arrow.factory import ArrowFactory
 # internal default factory.
 _factory = ArrowFactory()
 
+# FIXME Put the docs here, not in the factory
 
 def get(*args, **kwargs):
-    ''' Implements the default :class:`ArrowFactory <arrow.factory.ArrowFactory>`
-    ``get`` method.
+    ''' Calls the default :class:`ArrowFactory <arrow.factory.ArrowFactory>` ``get`` method.
 
     '''
 
     return _factory.get(*args, **kwargs)
 
 def utcnow():
-    ''' Implements the default :class:`ArrowFactory <arrow.factory.ArrowFactory>`
-    ``utcnow`` method.
+    ''' Calls the default :class:`ArrowFactory <arrow.factory.ArrowFactory>` ``utcnow`` method.
 
     '''
 
@@ -32,8 +31,7 @@ def utcnow():
 
 
 def now(tz=None):
-    ''' Implements the default :class:`ArrowFactory <arrow.factory.ArrowFactory>`
-    ``now`` method.
+    ''' Calls the default :class:`ArrowFactory <arrow.factory.ArrowFactory>` ``now`` method.
 
     '''
 
@@ -51,5 +49,5 @@ def factory(type):
     return ArrowFactory(type)
 
 
+# FIXME iso?
 __all__ = ['get', 'utcnow', 'now', 'factory', 'iso']
-
--- a/arrow/api.py
+++ b/arrow/api.py
@@ -13,7 +13,6 @@ from arrow.factory import ArrowFactory
 # internal default factory.
 _factory = ArrowFactory()
 
-# FIXME Put the docs here, not in the factory
 
 def get(*args, **kwargs):
     ''' Calls the default :class:`ArrowFactory <arrow.factory.ArrowFactory>` ``get`` method.
@@ -49,5 +48,4 @@ def factory(type):
     return ArrowFactory(type)
 
 
-# FIXME iso?
-__all__ = ['get', 'utcnow', 'now', 'factory', 'iso']
+__all__ = ['get', 'utcnow', 'now', 'factory']
--- a/arrow/api.py
+++ b/arrow/api.py
@@ -49,3 +49,4 @@ def factory(type):
 
 
 __all__ = ['get', 'utcnow', 'now', 'factory']
+
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -47,7 +47,6 @@ class Arrow(object):
         <Arrow [2013-05-05T12:30:45+00:00]>
 
     '''
-    # FIXME Why not accept the same set of timezone expressions everywhere?
 
     resolution = datetime.resolution
 
@@ -103,7 +102,7 @@ class Arrow(object):
         :param timestamp: an ``int`` or ``float`` timestamp, or a ``str`` that converts to either.
         :param tzinfo: (optional) a ``tzinfo`` object.  Defaults to local time.
 
-        Timestamps should always be UTC. If you have a non-UTC timestamp:
+        Timestamps should always be UTC. If you have a non-UTC timestamp::
 
             >>> arrow.Arrow.utcfromtimestamp(1367900664) .replace(tzinfo='US/Pacific')
             <Arrow [2013-05-07T04:24:24-07:00]>
@@ -139,7 +138,7 @@ class Arrow(object):
         :param dt: the ``datetime``
         :param tzinfo: (optional) a ``tzinfo`` object.  Defaults to dt.tzinfo, or UTC if naive.
 
-        If you only want to replace the timezone of naive datetimes:
+        If you only want to replace the timezone of naive datetimes::
 
             >>> dt
             datetime.datetime(2013, 5, 5, 0, 0, tzinfo=tzutc())
@@ -204,8 +203,6 @@ class Arrow(object):
         return the entire range.  Call with ``limit`` alone to return a maximum # of results from
         the start.  Call with both to cap a range at a maximum # of results.
 
-        **NOTE**: Unlike Python's ``range``, ``end`` **may** be included in the returned list.
-
         Supported frame values: year, quarter, month, week, day, hour, minute, second.
 
         Recognized datetime expressions:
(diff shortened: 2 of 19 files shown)


arrow/api.py
```diff
@@ -51,5 +48,4 @@ def factory(type):
     return ArrowFactory(type)
 
 
-__all__ = ['get', 'utcnow', 'now', 'factory', 'iso']
-
+__all__ = ['get', 'utcnow', 'now', 'factory']
```
Syeberman: `iso` doesn't exist any more, and was causing a warning when generating the docs.

arrow/arrow.py
```diff
@@ -118,11 +132,19 @@ def utcfromtimestamp(cls, timestamp):
...
-        :param tzinfo: (optional) a ``tzinfo`` object.  Defaults to UTC.
+        :param tzinfo: (optional) A :ref:`timezone expression <tz-expr>`.  Defaults to ``dt``'s
+            timezone, or UTC if naive.
+
+        If you only want to replace the timezone of naive datetimes::
+
+            >>> dt
+            datetime.datetime(2013, 5, 5, 0, 0, tzinfo=tzutc())
+            >>> arrow.Arrow.fromdatetime(dt, dt.tzinfo or 'US/Pacific')
+            <Arrow [2013-05-05T00:00:00+00:00]>
```
Syeberman: I figured this would be a common enough need that it should be documented.  It also highlights that `tzinfo` _always_ replaces the timezone.

arrow/arrow.py
```diff
@@ -88,11 +96,17 @@ def utcnow(cls):
...
+        ''' Constructs an :class:`Arrow <arrow.arrow.Arrow>` object from a timestamp, converted to
+        the given timezone.
 
         :param timestamp: an ``int`` or ``float`` timestamp, or a ``str`` that converts to either.
         :param tzinfo: (optional) a ``tzinfo`` object.  Defaults to local time.
 
+        Timestamps should always be UTC. If you have a non-UTC timestamp::
+
+            >>> arrow.Arrow.utcfromtimestamp(1367900664).replace(tzinfo='US/Pacific')
+            <Arrow [2013-05-07T04:24:24-07:00]>
```
Syeberman: This helps to highlight that `tzinfo` replaces the timezone.

arrow/arrow.py
```diff
@@ -165,46 +189,54 @@ def strptime(cls, date_str, fmt, tzinfo=None):
...
 
             - An :class:`Arrow <arrow.arrow.Arrow>` object.
             - A ``datetime`` object.
 
-        Recognized timezone expressions:
-
-            - A ``tzinfo`` object.
-            - A ``str`` describing a timezone, similar to 'US/Pacific', or 'Europe/Berlin'.
-            - A ``str`` in ISO-8601 style, as in '+07:00'.
-            - A ``str``, one of the following:  'local', 'utc', 'UTC'.
```
Syeberman: Rather than duplicating this list everywhere, it's now linked to via `:ref:`timezone expression <tz-expr>``.

arrow/arrow.py
```diff
@@ -165,46 +189,54 @@ def strptime(cls, date_str, fmt, tzinfo=None):
...
-
-        Usage:
+        Usage::
 
             >>> start = datetime(2013, 5, 5, 12, 30)
             >>> end = datetime(2013, 5, 5, 17, 15)
             >>> for r in arrow.Arrow.range('hour', start, end):
-            ...     print repr(r)
+            ...     print(repr(r))
```
Syeberman: Python 2 and 3 compatible

arrow/arrow.py
```diff
@@ -165,46 +189,54 @@ def strptime(cls, date_str, fmt, tzinfo=None):
...
             ...
             <Arrow [2013-05-05T12:30:00+00:00]>
             <Arrow [2013-05-05T13:30:00+00:00]>
             <Arrow [2013-05-05T14:30:00+00:00]>
             <Arrow [2013-05-05T15:30:00+00:00]>
             <Arrow [2013-05-05T16:30:00+00:00]>
 
+        **NOTE**: Unlike Python's ``range``, ``end`` *may* be included in the returned list::
```
Syeberman: This was a behaviour that tripped me up the first first times.

arrow/arrow.py
```diff
@@ -279,11 +314,6 @@ def span_range(cls, frame, start, end, tz=None, limit=None):
     # representations
 
     def __repr__(self):
-
-        dt = self._datetime
-        attrs = ', '.join([str(i) for i in [dt.year, dt.month, dt.day, dt.hour, dt.minute,
-            dt.second, dt.microsecond]])
```
Syeberman: This code was unused: I presumed it was safe to delete.

arrow/arrow.py
```diff
@@ -373,31 +406,25 @@ def replace(self, **kwargs):
...
-        Use plural property names to shift their current value relatively:
-
-        >>> arw.replace(years=1, months=-1)
-        <Arrow [2014-04-11T22:27:34.787885+00:00]>
+            >>> import arrow
+            >>> arw = arrow.utcnow()
+            >>> arw
+            <Arrow [2013-05-11T22:27:34.787885+00:00]>
+            >>> arw.replace(year=2014, month=6)
+            <Arrow [2014-06-11T22:27:34.787885+00:00]>
```
Syeberman: I made the indentation consistent in this docstring.

arrow/arrow.py
```diff
@@ -767,12 +788,13 @@ def time(self):
...
     def timetz(self):
-        ''' Returns a ``time`` object with the same hour, minute, second, microsecond and tzinfo. '''
+        ''' Returns a ``time`` object with the same hour, minute, second, microsecond and
+        tzinfo. '''
 
         return self._datetime.timetz()
 
     def astimezone(self, tz):
-        ''' Returns a ``datetime`` object, adjusted to the specified tzinfo.
+        ''' Returns a ``datetime`` object, converted to the specified timezone.
```
Syeberman: I tried to keep the same words used in index.rst: replace vs convert.  So I changed "adjusted" here for consistency.

arrow/factory.py
```diff
@@ -44,15 +44,16 @@ def get(self, *args, **kwargs):
         **None** to also get current UTC time::
 
             >>> arrow.get(None)
-            <Arrow [2013-05-08T05:51:43.316458+00:00]>
+            <Arrow [2013-05-08T05:51:49.016458+00:00]>
```
Syeberman: ...just so that this "now" isn't the same as the previous "now".

arrow/factory.py
```diff
@@ -64,14 +65,14 @@ def get(self, *args, **kwargs):
...
-            <Arrow [2013-05-08T05:54:34+0struct_time0:00]>
+            <Arrow [2013-05-08T05:54:34+00:00]>
 
         **One** ISO-8601-formatted ``str``, to parse it::
 
             >>> arrow.get('2013-09-29T01:26:43.830580')
             <Arrow [2013-09-29T01:26:43.830580+00:00]>
 
-        **One** ``tzinfo``, to get the current time in that timezone::
+        **One** ``tzinfo``, to get the current time **converted** to that timezone::
```
Syeberman: Here, the timezone is a conversion, but it is a replacement everywhere else in `get`.  I find this confusing and wonder if this particular case should be deprecated in favour of just calling `arrow.now(tz)` directly.

docs/conf.py
```diff
@@ -64,7 +64,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = ['_build']
+exclude_patterns = ['_build', '_themes/**README.rst']
```
Syeberman: The changes in this file are just to fix some warnings.

docs/index.rst
```diff
@@ -13,12 +13,12 @@ Arrow is heavily inspired by `moment.js <https://github.com/timrwood/moment>`_ a
 ----
 Why?
 ----
-Python's standard library and some other low-level modules have near-complete date, time and time zone functionality but don't work very well from a usability perspective:
+Python's standard library and some other low-level modules have near-complete date, time and timezone functionality but don't work very well from a usability perspective:
```
Syeberman: There was inconsistent use of both "time zone" and "timezone".  I personally prefer the latter, but I can change this.  (Even Python's own docs are inconsistent with this spelling.)

docs/index.rst
```diff
@@ -115,19 +115,19 @@ Create from timestamps (ints or floats, or strings that convert to a float):
...
 
 .. code-block:: python
 
     >>> arrow.get(datetime.utcnow())
     <Arrow [2013-05-07T04:24:24.152325+00:00]>
 
-    >>> arrow.get(datetime.now(), 'US/Pacific')
-    <Arrow [2013-05-06T21:24:32.736373-07:00]>
+    >>> arrow.get(datetime(2013, 5, 5), 'US/Pacific')
+    <Arrow [2013-05-05T00:00:00-07:00]>
```
Syeberman: I changed these examples because I can't think of a use case where you would want to replace the timezone for what is clearly a local time (`now`, that is).

andrewelkins:
Fantastic PR and docs to go along with it. Thank you.

PR #344, opened 2016-07-12
return NotImplemented instead of raising TypeError

flokli:
`TypeError` will already be raised by Python, if both object
implementations  don't know how to handle the other type.

This allows other classes to implement a `__rOperation__`, even
when arrow doesn't know how to handle it with `__Operation__` itself.

From the docs:

**NotImplemented:**
Special value which should be returned by the binary special methods
(e.g. `__eq__()`, `__lt__()`, `__add__()`, `__rsub__()`, etc.) to indicate that
the operation is not implemented with respect to the other type; may be
returned by the in-place binary special methods (e.g. `__imul__()`,
`__iand__()`, etc.) for the same purpose. Its truth value is `true`.

**Binary arithmetic operations:**
[...]
If one of those methods does not support the operation with the supplied
arguments, it should return `NotImplemented`.

**Rich comparison methods:**
A rich comparison method may return the singleton `NotImplemented` if it
does not implement the operation for a given pair of arguments.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -681,7 +681,7 @@ class Arrow(object):
         if isinstance(other, (timedelta, relativedelta)):
             return self.fromdatetime(self._datetime + other, self._datetime.tzinfo)
 
-        raise TypeError()
+        return NotImplemented
 
     def __radd__(self, other):
         return self.__add__(other)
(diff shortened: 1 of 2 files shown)


flokli:
Python 2 doesn't raise `TypeError` if binary arithmetic operations return `NotImplemented`, so I fixed this in __cmp__.

andrewelkins:
On first review this looks okay, but I'll need to do some real world testing.

flokli:
@andrewelkins: Great! Tell me in case you need any more info :-)

Syeberman:
There were some merge conflicts, but the branch wasn't open for edits, so I've continued the work at https://github.com/crsmithdev/arrow/pull/522.  Thanks @flokli, and sorry for the delay!

PR #358, opened 2016-09-11
Rescale large timestamps automatically

### LINKED ISSUE #357: Parse timestamp with milliseconds/microseconds
Arrow version 0.8.0
arrow.get cannot handle large timestamps because datetime throws a ValueError when a timestamp results in a year greater than or equal to 10000.

``` python

>>> import arrow
>>> arrow.get(1000000000000)
  File "/usr/local/lib/python2.7/site-packages/arrow/api.py", line 23, in get
  File "/usr/local/lib/python2.7/site-packages/arrow/factory.py", line 144, in get
  File "/usr/local/lib/python2.7/site-packages/arrow/arrow.py", line 114, in utcfromtimestamp
ValueError: year is out of range
```

However,  most of the time, that error stems from the fact that the timestamp is not in seconds but in ms or ns.

I'm suggesting that if the timestamp is too large for datetime to parse properly, that means it's not in seconds and we should divide it appropriately.

What do you think?

henridwyer:
Fixes #357

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -904,7 +904,13 @@ class Arrow(object):
     def _get_timestamp_from_input(timestamp):
 
         try:
-            return float(timestamp)
+            timestamp = float(timestamp)
+            if timestamp >= 253402300800000: # nanoseconds
+                return timestamp / 1000000
+            elif timestamp >= 253402300800: # milliseconds
+                return timestamp / 1000
+            else:
+                return timestamp
         except:
             raise ValueError('cannot parse \'{0}\' as a timestamp'.format(timestamp))
 
--- a/tests/factory_tests.py
+++ b/tests/factory_tests.py
@@ -58,6 +58,17 @@ class GetTests(Chai):
         with assertRaises((OverflowError, ValueError,)):
             self.factory.get(timestamp)
 
+        # Issue 357 - Large timestamps that can't be in seconds
+        timestamp = 1000000000000. # this needs to be treated as ms
+        timestamp_dt = datetime.utcfromtimestamp(timestamp / 1000).replace(tzinfo=tz.tzutc())
+        
+        assertEqual(self.factory.get(timestamp), timestamp_dt)
+        
+        timestamp = 1000000000000000. # this needs to be treated as us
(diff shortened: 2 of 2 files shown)


arrow/arrow.py
```diff
@@ -904,7 +904,13 @@ def _get_iteration_params(cls, end, limit):
     def _get_timestamp_from_input(timestamp):
 
         try:
-            return float(timestamp)
+            timestamp = float(timestamp)
+            if timestamp >= 253402300800000: # nanoseconds
+                return timestamp / 1000000
+            elif timestamp >= 253402300800: # milliseconds
```
bukzor: What does this number represent?

It seems to me that the timestamp `1000` is fully ambiguous with respect to units (does it mean 1000 seconds since 1970 or 1000 nanoseconds?). Since we can't infer the units for all values, it seems futile (invalid) to do it for some values.

henridwyer: 1000 is ambiguous, no doubt.
However anything above 253402300800 (which corresponds to 1/1/10000), datetime is not designed to handle. Therefore, anything above that can't be a timestamp in seconds, so arrow could assume it's not rather than throwing an error.

andrewelkins:
An interesting thought, but I don't like hidden manipulation. If the timestamp isn't in a correct format it should be handled before getting to Arrow.

awbacker:
I'm very interested in this, since we have a mix of these coming in.  For the moment I've solved it by just checking ` > 2**32`.  Probably a better way, but it would be great to have canonical handling of this, particularly since JS `Date.now()` returns the long one.

```
def ts_to_arrow(value):
    assert isinstance(value, six.integer_types)
    return arrow.get(value / 1e3) if value > 2**32 else arrow.get(value)
```

AlJohri:
@andrewelkins how about with @awbacker's approach?

andrewelkins:
Sure that makes sense

matrix0415:
Is it possible for merging this PR? thanks!

jadchaar:
Closing this PR in favor of this updated PR: https://github.com/crsmithdev/arrow/pull/615.

PR #366, opened 2016-10-20
support nanoseconds in RFC 3339 format

tangyouze:


bukzor:
I've fixed the more general case in https://github.com/crsmithdev/arrow/pull/372

andrewelkins:
This is fixed by #372

PR #371, opened 2016-11-15
Add support for universal wheels

aragilar:
Since arrow is a pure python module which support python 2 and 3 with the same code, it can be put in a universal wheel.


bukzor:
shipit 👍

andrewelkins:
Looks good to me.

PR #372, opened 2016-11-16
Support gnu date iso 8601

bukzor:
There's two things that arrow isn't doing quite right with respect to fractional seconds:

 1. The separator is sometimes comma. In fact, this is the "preferred" separator per the standard.
 2. The fractional seconds have arbitrary precision; the only requirement is that the sender and receiver agree on the precision. For example, a high-energy physics lab might use iso-8601 timestamps with attosecond precision without violating the standard.

I've left a couple TODO's below, on which I'd like to get your input. I had to make a couple compromises when implementing this:

 1. Python's stdlib datetime only accepts microseconds, so any fractional seconds of higher precision have some data loss, but are parsed as accurately as possible. 
2. In "fixing" the date-format code, I made any number of S match any number of fractional seconds. I can't see any drawback to this, and some benefit, but you may consider this to be incorrect.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,9 @@
 [run]
 branch = True
-source = arrow
+source =
+    tests
+    arrow
+
+[report]
+show_missing = True
+fail_under = 100
--- a/arrow/api.py
+++ b/arrow/api.py
@@ -51,5 +51,5 @@ def factory(type):
     return ArrowFactory(type)
 
 
-__all__ = ['get', 'utcnow', 'now', 'factory', 'iso']
+__all__ = ['get', 'utcnow', 'now', 'factory']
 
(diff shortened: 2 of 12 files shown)


andrewelkins:
I like it. The one thing I'd like is to remove the rounding bit. That seems like a bit of magic from the end users point of view. Or if it is now a "feature" the docs need to be updated to reflect it.

bukzor:
I've removed the rounding, but left it in the history in case anyone wants to discuss or look at it in future.

bukzor:
The last commit adds the rounding, as well as its documentation and necessary test changes.

bukzor:
Bump

jamesflinn:
👍

kdeal:
:shipit:

andrewelkins:
Excellent. I'll get this merged in today with the rest of the ones I've tagged.

bukzor:
Great! Let me know when it gets to pypi.

bukzor:
(bump)

PR #377, opened 2016-11-28
Fix negative offset for UTC

mrmrwat:
Specifying ZZ in a format string gives an offest of -00:00 for UTC,
despite the arrow object repr showing +00:00. Detect `using >= 0`
for consistency.


1CM69:
Cheers for looking at this

mrmrwat:
Hmmm, the tests failed because of the format change. I guess either they want it to be like that, or they're locking in the behaviour for some other reason. I'll fix the tests, but there's no guarantee that the maintainers will appreciate the patch! 😄

andrewelkins:
I'll merge. This goes against what moment.js does and just is odd behavior. Thanks for the PR

PR #379, opened 2016-11-29
Fixes #378: __getattr__ has an off-by-one error in its quarter handling

### LINKED ISSUE #378: __getattr__ doesn't compute quarters correctly
Arrow's quarter start date is one month too early. For example, it thinks 1 March is part of Q2:

```py
In [2]: dt = arrow.get("2016-03-01")

In [3]: dt.month
Out[3]: 3

In [4]: dt.quarter
Out[4]: 2
```

This is tested on 0.8.0.

ns-cweber:
Fixes issue https://github.com/crsmithdev/arrow/issues/378

I added test cases for quarter end dates to make sure they were being included in the correct quarter (as well as a fix).

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -310,7 +310,7 @@ class Arrow(object):
             return self.isocalendar()[1]
 
         if name == 'quarter':
-            return int(self.month/self._MONTHS_PER_QUARTER) + 1
+            return int((self.month-1)/self._MONTHS_PER_QUARTER) + 1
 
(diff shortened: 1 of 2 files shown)


ns-cweber:
@andrewelkins Do you happen to know when the next release will be? Is there some sort of release schedule I could look at?

andrewelkins:
No release schedule. I've updated the version and should push version 0.10.0 today or tomorrow to pypi.

PR #383, opened 2016-12-02
Dehumanize implementation and tests

jasongwkim:
Original issue: https://github.com/crsmithdev/arrow/issues/360
This is a baseline implementation of dehumanize, can definitely expand functionality later.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -11,6 +11,7 @@ from datetime import datetime, timedelta, tzinfo
 from dateutil import tz as dateutil_tz
 from dateutil.relativedelta import relativedelta
 import calendar
+import copy
 import sys
 import warnings
 
@@ -663,7 +664,6 @@ class Arrow(object):
             <Arrow [2013-05-05T12:29:43+00:00]>
 
         '''
-        import copy #put this on top of arrow.py
         dtobj = copy.deepcopy(self._datetime)
         
         temp = tstring.split(' ')
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -649,7 +649,10 @@ class Arrow(object):
             Each b_k represents a string representing the unit of time measurement
             c is the string 'ago' or 'later'
             i.e. dehumanize('13 months 2 days 3 months 1 year ago') would return an arrow object representing
-            2 years, 4 months, and 2 days previous from the current time
+            2 years, 4 months, and 2 days previous from the current time. It is calculated using the differences
+            of the least significant units first, i.e. '12 months 3 years 365 days later' from 2011-01-01 would
+            return an arrow object representing 2016-01-01 but from 2012-01-01 would return 2016-12-31 due to the
+            leap day.
 
         Usage::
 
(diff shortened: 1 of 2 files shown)


andrewelkins:
Just need to review this and then I will merge.

tests/arrow_tests.py
```diff
@@ -1221,6 +1221,129 @@ def test_years(self):
...
+        result = arw.dehumanize('3 year 12 months 365 days ago')
+        assertEqual(result, arrow.Arrow(2008,1,2,0,0,0))
+
+        result = arw2012.dehumanize('3 year 12 months 365 days ago')
+        assertEqual(result, arrow.Arrow(2007,1,1,0,0,0))
+
+    def test_errors(self):
+        arw = arrow.Arrow.fromdatetime(self.datetime)
+
+        test = True
```
laixintao: You can't test with try-catch, since you don't know which Exception will be raisen.

use assertRaise instead.

```
class MyTestCase(unittest.TestCase):
    def test1(self):
        self.assertRaises(SomeCoolException, mymod.myfunc)
```

see also https://docs.python.org/dev/library/unittest.html#unittest.TestCase.assertRaises

arrow/arrow.py
```diff
@@ -640,6 +641,86 @@ def format(self, fmt='YYYY-MM-DD HH:mm:ssZZ', locale='en_us'):
...
+            >>> import arrow
+            >>> t = arrow.Arrow(2013, 5, 5, 12, 30, 45)
+            <Arrow [2013-05-05T12:30:45+00:00]>
+            >>> t.dehumanize('2 seconds 1 minute ago')
+            <Arrow [2013-05-05T12:29:43+00:00]>
+
+        '''
+        dtobj = copy.deepcopy(self._datetime)
+        
+        temp = tstring.split(' ')
```
laixintao: temp is not a good variable name, use times, time_strs, or something else

arrow/arrow.py
```diff
@@ -640,6 +641,86 @@ def format(self, fmt='YYYY-MM-DD HH:mm:ssZZ', locale='en_us'):
...
+            raise ValueError("invalid input of time")
+        
+        months = 0
+        years = 0
+        
+        for i in range(0,len(times),2):
+            tval = int(times[i])
+            tunit = times[i+1]
+            
+            if tunit == 'second' or tunit == 'seconds':
```
laixintao: `if tunit in ['second', 'seconds']:` is more flexible

rautamiekka: A tuple is faster to build than a list, though.

systemcatch:
Would be nice to add this feature and expand on it later. No conflicts with with master which is lucky.

@jasongwkim I know this PR is almost 2 years old but if you could implement some of @laixintao's suggestions we can get this moving.

systemcatch:
Thanks for the PR, this is an interesting proposition that we'd like to see in the arrow library some day. Since this is so old I'm going to close it now. New PR's or implementations welcome.

Issue #355, opened 2016-08-31
Specify start date in span('week')


mehtadev17:
arrow.utcnow().span('week') starts on Monday and ends on Sunday. I would like an option to start the week on Sunday and end on Saturday.

byin0317:
I am interested in working on this issue, will look further into by next week or so

systemcatch:
We could use https://docs.python.org/3.8/library/calendar.html#calendar.Calendar for this.

sawhitt:
I am interested in working on this, can I look into it?

oakt:
Can we work on this?

jadchaar:
Hey @oakt, if you’d like to contribute this change, PRs are welcome :).

yiransii:
@jadchaar 
Can I take this?

PR #387, opened 2016-12-05
Adjustable Start Index for Spans (Per Issue #355)

wtreid:
This commit addresses git issue 355 (https://github.com/crsmithdev/arrow/issues/355).  In addition to providing the ability to give a start index for a weekday, it also allows a user to give a start index for all spans but quarters.  For example, you could create a span of a month starting on the 7th of the month.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -527,7 +527,7 @@ class Arrow(object):
         return self.__class__(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
             dt.microsecond, dt.tzinfo)
 
-    def span(self, frame, count=1, start_index=1):
+    def span(self, frame, count=1, start_index=0):
         ''' Returns two new :class:`Arrow <arrow.arrow.Arrow>` objects, representing the timespan
         of the :class:`Arrow <arrow.arrow.Arrow>` object in a given timeframe.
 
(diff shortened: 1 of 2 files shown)


andrewelkins:
Thanks for the extensive PR. I'll review shortly.

rkunnamp:
This would be highly useful. Thanks.

cadupont:
That would be super useful. @andrewelkins Is there anything we can do to speed up the process?
I'm quite interested by this update. Thanks

Syeberman:
@wtreid @andrewelkins I have an idea to make this even more flexible: rename `start_index` to `shift`, and accept a dict of keywords as per the `shift()` function:

```python
arrow.utcnow().span('hour', shift={'minutes': 5})
```

This allows for, as an example, shifts of seconds when the span is an hour.  It's self-documenting, as the API and docs already define what the "shift" concept means.  It also greatly simplifies the PR, because the giant `if` statement becomes `floor = floor.shift(**shift)`.

systemcatch:
@wtreid sorry for the delay on this. If we can fix the conflicts then use @Syeberman's suggestion then this can be merged.

systemcatch:
This is a good idea but using @Syeberman's solution seems like the more elegant approach. If you'd like to re-implement this feel free to open another PR, but otherwise we hope to implement this in the future.

PR #391, opened 2016-12-07
Added support for ZZZ while formatting

### LINKED ISSUE #390: ZZZ parser will not format back
While the docs describes that [tokens](http://arrow.readthedocs.io/en/latest/#tokens) can be used both for parsing and formatting, the "new" `ZZZ` token will not work for formatting:

```
>>> import arrow
>>> arrow.now().format('ZZZ')
'+00:00+0000'
```

rafaelsierra:
This fixes #390, although I am not entirely happy that `ZZZ` will parse `America/Sao_Paulo` and format as `BRT`, I could not find a reliable way to return `America/Sao_Paulo` from things like `-0300` and `BRT`.

### DIFF
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -9,7 +9,7 @@ from arrow import util, locales
 
 class DateTimeFormatter(object):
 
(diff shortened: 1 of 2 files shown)


xrisk:
Merge this please or update the docs to that effect!

systemcatch:
I think this can be merged ok without issues. Thanks for the contribution @rafaelsierra, sorry it took so long to be included.

PR #349, opened 2016-08-12
Separate replace & shift function

Syeberman:
This resolves the merge conflicts from https://github.com/crsmithdev/arrow/pull/276.  It also fixes (in https://github.com/crsmithdev/arrow/pull/349/commits/b5eee6ddf91f6479309a8597f3052a0f8ffa37a0) a bug in the `quarters` functionality, adds `quarters` to `shift`, and gets back to 100% code coverage.

Separating replace and shift functionality would save a lot of headaches!

PR #406, opened 2017-01-04
Replace ".replace" with new ".shift" function in docs

samdroid-apps:
Commit 15d903a (from #349), separated the replace and shift features
into two functions.  However the docs still show the old, deprecated and
more confusing use of replace to shift the time.


rsyring:
I came to create an issue for this only to found it was already done.  Thanks!

PR #407, opened 2017-01-05
Added more human-readable error message for invalid frames for range/span

wonderb0lt:
This is a pretty minor change, I just goofed it up and was a bit confounded by the non-descriptive error message. This should make it clearer what happened.


andrewelkins:
Looks good, but can you fix it so it passes all tests? https://travis-ci.org/crsmithdev/arrow/jobs/189177908

wonderb0lt:
Yup, gotcha!

PR #417, opened 2017-02-02
A short duration isn't necessarily "just now"

### LINKED ISSUE #416: .humanize for near-zero duration says "just now"
This is very confusing for a field that is meant to be duration.
Consider, I may have a very short duration event that occurred weeks ago.

Arrow will say this event was "just now".
A more accurate description would be "instantly".

bukzor:
Closes #416

obviously the other locales need to be updated to match, but this is still an improvement

### DIFF
--- a/arrow/locales.py
+++ b/arrow/locales.py
@@ -192,7 +192,7 @@ class EnglishLocale(Locale):
     future = 'in {0}'
 
     timeframes = {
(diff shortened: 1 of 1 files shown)


systemcatch:
Hey @bukzor I know it's been a long time but if you fix the broken tests we can get this merged.

PR #429, opened 2017-04-17
Add 3.6, pypy, and pypy3, and refactor CI integration

medecau:
This pull-request:

- adds python3.6, pypy, and pypy3 to the tests environment list
- moves dependency definition into setup.py while remaining sensible to python 2.6 requirements
- change travis config to use tox

If this pull-request is accepted it'll require changes to the Makefile. Project maintainer have access to this branch and can push changes before merging with master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,14 @@ python:
   - 3.3
   - 3.4
   - 3.5
+  - 3.6
 install:
     - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then make build26; fi
     - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then make build27; fi
     - if [[ $TRAVIS_PYTHON_VERSION == '3.3' ]]; then make build33; fi
     - if [[ $TRAVIS_PYTHON_VERSION == '3.4' ]]; then make build34; fi
     - if [[ $TRAVIS_PYTHON_VERSION == '3.5' ]]; then make build35; fi
+    - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then make build36; fi
     - pip install codecov
 script: make test
 after_success: codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,27 @@
 sudo: false
 language: python
-python:
-  - 2.6
-  - 2.7
-  - 3.3
-  - 3.4
-  - 3.5
-  - 3.6
+matrix:
+  include:
+    - python: 2.6
+      env: TOXENV=py26
+    - python: 2.7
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -25,6 +26,15 @@ def grep(attrname):
 
 file_text = read(fpath('arrow/__init__.py'))
 
+install_requires = ['backports.functools_lru_cache',
+                    'python-dateutil',
+                    'simplejson']
```
Syeberman: While Arrow is compatible with simplejson, it is not a requirement for installation.  Please move `'simplejson'` back to tox.ini.

setup.py
```diff
@@ -25,6 +26,15 @@ def grep(attrname):
 
 file_text = read(fpath('arrow/__init__.py'))
 
+install_requires = ['backports.functools_lru_cache',
```
Syeberman: `backports.functools_lru_cache` is only required for Python 2.x.  Could you do a similar `install_requires.append` as you do for chai?

.travis.yml
```diff
@@ -1,17 +1,27 @@
...
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.3' ]]; then make build33; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.4' ]]; then make build34; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.5' ]]; then make build35; fi
-    - pip install codecov
-script: make test
-after_success: codecov
+    - pip install tox
+script: tox
+after_success:
+  - pip install codecov
```
Syeberman: Does Travis CI cache the `install` step for future builds?  If so, we should install codecov in the install step.

medecau:
Thanks for reviewing.

I don't know when I will have time to add the necessary changes to this PR, busy with work and away from personal dev machine.

Alternatively, this PR accepts changes from maintainers of the main repo.

.travis.yml
hugovk: You can cache like this:
https://docs.travis-ci.com/user/caching/#pip-cache

setup.py
```diff
@@ -36,9 +46,7 @@ def grep(attrname):
     license='Apache 2.0',
     packages=['arrow'],
     zip_safe=False,
-    install_requires=[
-        'python-dateutil'
-    ],
+    install_requires=install_requires,
     test_suite="tests",
     classifiers=[
```
hugovk: Don't forget to add the new versions the classifiers. Can also add CPython/PyPy:

https://pypi.python.org/pypi?%3Aaction=list_classifiers

systemcatch:
Python 2.6/3.3 support has been dropped and 3.6 is now supported.

PR #430, opened 2017-04-17
Fix grammar of ArabicLocale

satiani:


satiani:
@crsmithdev @andrewelkins Please let me know if this PR follows the contribution guidelines. We'd really like to see it merged so we can get back to using the pypi version of arrow ASAP.

Syeberman:
Hi @satiani, thanks for the contribution!  (Sorry it took so long, but we've recently added new maintainers to the project and are working through the backlog.)

I have just one concern about this PR, and it's that it removes `ArabicLocale` completely, which will break any downstream code that references this class.  Would it make sense to rename `DefaultArabicLocale` back to `ArabicLocale`?

arrow/locales.py
```diff
@@ -1076,37 +1076,101 @@ class AzerbaijaniLocale(Locale):
...
+    def _format_timeframe(self, timeframe, delta):
+        form = self.timeframes[timeframe]
+        delta = abs(delta)
+        if isinstance(form, dict):
+            if delta == 2:
+                form = form['double']
+            elif delta > 2 and delta <= 10:
+                form = form['ten']
+            else:
+                form = form['higher']
```
Syeberman: Under this, a `delta` of 1 will get `form['higher']`.  Is that correct grammar?

satiani: The assumption is that if its >1 `timeframe` will be `days`, `minutes`, etc. If delta is 1 then `timeframe` would be `day`, `minute`, etc. and correct grammar will be chosen. That assumption seems also to be true for other Locale implementations, including English.

satiani:
I can definitely rename it. I see you merged the PR, thanks for that!
Should I open a new PR to rename the class?

arrow/locales.py
Syeberman: Makes sense. (I'm not as familiar with this area of the code.)

Syeberman:
I updated your branch with the latest master, to fix a conflict and retest.  The PR is still open, so let's make the change here.

satiani:
@Syeberman I renamed the class name.

Let me know if you have any other feedback. If there is no other feedback, do you have any estimate of when this version may be available on PyPi? I'd like to switch to using the PyPi version instead of my custom build.

Syeberman:
@satiani Merged!

As for the release to PyPI, I can't give a date just yet, but we are actively working to catch up to PRs and then update readthedocs.io and PyPI.

PR #440, opened 2017-06-17
Return generator instead of list in Arrow.range(), Arrow.span_range(), and Arrow.interval(). Fix #251.

### LINKED ISSUE #251: Add xrange method
Can you add xrange method to arrow object ?
It's like range method but it return an iterator.

Thank you

rominf:
Breaking change. Probably for the next major release.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -192,7 +192,7 @@ class Arrow(object):
 
     @classmethod
     def range(cls, frame, start, end=None, tz=None, limit=None):
-        ''' Returns a list of :class:`Arrow <arrow.arrow.Arrow>` objects, representing
+        ''' Returns a generator of :class:`Arrow <arrow.arrow.Arrow>` objects, representing
         an iteration of time between two inputs.
 
         :param frame: the timeframe.  Can be any ``datetime`` property (day, hour, minute...).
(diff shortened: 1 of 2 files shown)


arrow/arrow.py
```diff
@@ -352,10 +351,10 @@ def interval(cls, frame, start, end, interval=1, tz=None):
         if interval < 1:
             raise ValueError("interval has to be a positive integer")
 
-        spanRange = cls.span_range(frame,start,end,tz)
+        spanRange = list(cls.span_range(frame, start, end, tz))
 
         bound = (len(spanRange) // interval) * interval
-        return [ (spanRange[i][0],spanRange[i+ interval - 1][1]) for i in range(0,bound, interval) ]
+        return ((spanRange[i][0], spanRange[i + interval - 1][1]) for i in range(0, bound, interval))
```
Syeberman: You could keep `spanRange` a generator, and return a generator, with something like this (completely untested, likely missing some boundary cases):

```python
spanRange = cls.span_range(frame, start, end, tz)
while True:
    intvlStart, intvlEnd = next(spanRange)  # StopIteration when exhausted
    for _ in range(interval-1):
        _, intvlEnd = next(spanRange)  # StopIteration when exhausted
    yield intvlStart, intvlEnd
```

Syeberman:
Work will continue on https://github.com/crsmithdev/arrow/pull/458

PR #444, opened 2017-07-03
warning when use get with only one arg and assign kwargs.

laixintao:
get('2017-07-03', tzinfo='Asia/Shanghai') will ignore tzinfo
silently, which makes us very confused.
```

In [1]: from arrow.api import get

In [2]: get('2017-01-01')
Out[2]: <Arrow [2017-01-01T00:00:00+00:00]>

In [3]: get('2017-01-01', tzinfo='Asia/Shanghai')
Out[3]: <Arrow [2017-01-01T00:00:00+00:00]>  # Not what I expacted

In [4]: get('2017-01-01', 'YYYY-MM-DD', tzinfo='Asia/Shanghai')
Out[4]: <Arrow [2017-01-01T00:00:00+08:00]>
```

better make a warning when use like that, so user can choose weather delete tzinfo or
use get('2017-07-03').replace(tzinfo='Asia/Shanghai')

### DIFF
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -8,6 +8,8 @@ construction scenarios.
 
 from __future__ import absolute_import
 
+import warnings
+
 from arrow.arrow import Arrow
 from arrow import parser
 from arrow.util import is_timestamp, isstr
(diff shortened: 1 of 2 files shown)


Syeberman:
With https://github.com/crsmithdev/arrow/pull/516, `tzinfo` is now respected in this case.

Thanks for the PR @laixintao, and sorry for the delay.

PR #445, opened 2017-07-08
add nepali locale support

techgaun:
I'm adding Nepali locale support as part of this PR. thanks!


Syeberman:
Thanks for the contribution @techgaun!

PR #453, opened 2017-09-11
fix tzinfo support for arrow.Factory.get

yunfan:
the Factory method get currently accept tzinfo when args count equals 2, but when its only give 1 args, it dont accept tzinfo, which looks like a missing. so i fix that by adding the missed part `tzinfo=tz`

### DIFF
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -154,7 +154,10 @@ class ArrowFactory(object):
 
             # try (int, float, str(int), str(float)) -> utc, from timestamp.
             if is_timestamp(arg):
-                return self.type.utcfromtimestamp(arg)
+                if tz is not None:
+                    return self.type.fromtimestamp(arg)
+                else:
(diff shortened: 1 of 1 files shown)


Syeberman:
Hi @yunfan, thanks for your contribution!

The current behaviour is actually by design, [according to the docs](https://github.com/crsmithdev/arrow/blob/fd723b46a6d84ea75af329fe78da6afaf27dfeba/arrow/factory.py#L37):

> `tzinfo` replaces the timezone unless using an input form that is explicitly UTC or specifies the timezone in a positional argument.

I agree that this behaviour is surprising.  I'll be considering ways to better-document this method, and perhaps add some additional argument checking for better error messages.

In the meantime, you can always call `arrow.get(obj).replace(tzinfo=tz)` to explicitly set any timezone to a different value.

Thanks again!

yunfan:
ok thanks for the replying. but now i am worried about this project itself.

PR #458, opened 2017-09-23
Return generator instead of list in Arrow.range() et al

Syeberman:
Based off of @rominf's PR https://github.com/crsmithdev/arrow/pull/440, implementing https://github.com/crsmithdev/arrow/issues/251, and using an idea from @pganssle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,9 @@
 
 .coverage
 nosetests.xml
+tests/.noseids
 
+.venv/
 local/
 dist/
 docs/_build/
@@ -19,3 +21,6 @@ docs/_build/
 
 # tox
 .tox
+
+# VS Code
+.vscode/
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -191,9 +191,10 @@ class Arrow(object):
     # factories: ranges and spans
 
     @classmethod
+    @util.list_to_iter_deprecation
     def range(cls, frame, start, end=None, tz=None, limit=None):
-        ''' Returns a generator of :class:`Arrow <arrow.arrow.Arrow>` objects, representing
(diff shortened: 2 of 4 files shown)


pypingou:
Note that this is API breaking and if we want to follow the semantic versioning scheme http://semver.org/ will require bumping the major version.

Syeberman:
@pypingou All three modified methods are decorated with `@util.list_to_iter_deprecation`, so they will continue to return `lists`.  As such, this PR is not itself a breaking change, but when we go to 1.0 we can remove `@util.list_to_iter_deprecation` and return actual iterators from these methods.

Syeberman:
@pypingou @andrewelkins What do you think?  Should we move forward with this, or are the benefits not worth breaking compatibility?

rhan-mentad:
Personally I don't see the value of this for use cases at work but we are already having to do type conversion (tuple to list) in another project so list wrapping shouldn't be an issue for where we don't typeconvert.

In terms of api breaking what about adding an iterative version of this using a naming convention similar to `funcy`?

Eg `irange` -> iterative range
`range` -> non iterative range.

Syeberman:
@rhan-mentad Adding an `irange` is another possibility, but the general direction in Python has been to prefer returning iterators directly rather than having a separate "iterator version" of a function.  (`xrange` merged with `range`, `iteritems` merged with `items`, etc.)

It's so easy to just wrap `list` or `tuple` around a function call that returns an iterator.  Certainly, in newly created APIs, I would prefer to always return an iterator.  The decision here is whether it's worth converting this existing API.

Syeberman:
I'm going to merge this on Feb 1, 2018, unless there are objections.

PR #464, opened 2017-10-01
Support microsecond precision timestamp on Python 3.3+

zmwangx:
Previous implementation of `arrow.Arrow.timestamp()` has precision of a second, since `datetime.datetime.utctimetuple()` returns a `time.struct_time` which doesn't even have a microsecond field.

In this commit, we switch to [`datetime.datetime.timestamp()`][1], available on Python 3.3 or later, which supports microsecond precision.

By the way I'm using `hasattr` for feature test now. Not sure if you prefer try...except-styled feature test.

[1]: https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -424,7 +424,10 @@ class Arrow(object):
         ''' Returns a timestamp representation of the :class:`Arrow <arrow.arrow.Arrow>` object, in
         UTC time. '''
 
-        return calendar.timegm(self._datetime.utctimetuple())
+        if hasattr(self._datetime, 'timestamp'):
+            return self._datetime.timestamp()
+        else:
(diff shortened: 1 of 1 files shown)


zmwangx:
Hmm, how do I bring coverage to 100% when there's a branching based on feature availability?

Syeberman:
Arrow's `timestamp` specifically returns an `int` with second-precision.  Use the existing `float_timestamp` if you need microsecond precision.  Unfortunately, Arrow's `timestamp` predates `datetime.timestamp`, so it has diverged slightly...we'll need to consider how to clear up this ambiguity.

Thanks for the PR @zmwangx, it's always appreciated!

zmwangx:
Oops, I didn't realize there's a `float_timestamp`. Sorry.

PR #466, opened 2017-10-09
Fix arrow.Arrow.interval()

dwrpayne:
arrow.Arrow.interval() produces unexpected results if the interval doesn't evenly divide the range. It leaves off the last chunk of time.

```shell
$> start = datetime(2013, 5, 5, 12, 30)
$> end = datetime(2013, 5, 5, 17, 15)
$> for r in arrow.Arrow.interval('hour', start, end, 4): print (r)
(<Arrow [2013-05-05T12:00:00+00:00]>, <Arrow [2013-05-05T15:59:59.999999+00:00]>)
```

Expected result:
```shell
$> for r in arrow.Arrow.interval('hour', start, end, 4): print (r)
(<Arrow [2013-05-05T12:00:00+00:00]>, <Arrow [2013-05-05T15:59:59.999999+00:00]>)
(<Arrow [2013-05-05T16:00:00+00:00]>, <Arrow [2013-05-05T17:59:59.999999+00:00]>)
```

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -351,14 +351,11 @@ class Arrow(object):
         '''
         if interval < 1:
             raise ValueError("interval has to be a positive integer")
-
-        spanRange = cls.span_range(frame,start,end,tz)
-
-        bound = (len(spanRange) // interval) * interval
-        _range = [ (spanRange[i][0],spanRange[i+ interval - 1][1]) for i in range(0,bound, interval) ]
-        
-        if (bound < len(spanRange)):
-            _range.append((spanRange[bound][0], spanRange[-1][1]))
+        range_span = cls.span_range(frame, start, end, tz)
+        bound = (len(range_span) // interval) * interval
+        _range = [(range_span[i][0], range_span[i + interval - 1][1]) for i in range(0,bound, interval) ]        
+        if bound < len(range_span):
+            _range.append((range_span[bound][0], range_span[-1][1]))
         return _range
 
     # representations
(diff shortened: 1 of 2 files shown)


andrewelkins:
Can you add a test case for this?

dwrpayne:
Sure, I'll do that tomorrow. Can you explain why it failed the CI build? I
can't find a full log that details exactly which text didn't pass.

rhan-mentad:
Imho it would be better the last chunk of time be in a separate interval rather than appended to the last time's end date. That way you are guaranteed the maximum time span between any two dates is the specified interval rather than the last one potentially being higher.

rhan-mentad:
can also fix some pep8 issues along the way with this PR ;)
```
if interval < 1:
        raise ValueError("interval has to be a positive integer")
    range_span = arrow.Arrow.span_range(frame, start, end, tz)
    bound = (len(range_span) // interval) * interval
    _range = [(range_span[i][0], range_span[i + interval - 1][1]) for i in range(0, bound, interval)]
    if bound < len(range_span):
        _range.append((range_span[bound][0], range_span[-1][1]))
    return _range
```

dwrpayne:
I'm confused. I wrote it so that if, for example, you ask for intervals of 1 hour over a 4.5 hour range, you are returned a list of length 5. The first four elements have a 1 hour interval and the 5th has the extra half hour.

dwrpayne:
The original implementation just dropped the final half hour.

dwrpayne:
@rhan-mentad I think you misunderstood. I'm not appending the extra chunk of time to the last pair, I'm appending it to the return list.

Interval() returns a list of (start,end) tuples. I'm just appending a final (start,end) tuple in cases where there is a bit of time in the range left over.

PR #467, opened 2017-10-13
Store locale in Arrow object so it can be used by format

eezomancer:
https://github.com/crsmithdev/arrow/issues/309

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -698,7 +698,7 @@ class Arrow(object):
             '2013-05-09 03:56:47 -00:00'
 
         '''
-        locale = self.locale if local is None and hasattr(self, 'locale') else 'en_us'
+        locale = self.locale if locale is None and hasattr(self, 'locale') else 'en_us'
         return formatter.DateTimeFormatter(locale).format(self._datetime, fmt)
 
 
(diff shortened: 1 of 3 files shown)


Syeberman:
Hi @eezomancer, thanks for the contribution!

I feel this can be accomplished a different way that wouldn't require changes to Arrow.  Two options come to mind, either creating a subclass in your project:

```python
class FrCaArrow(arrow.Arrow):
  def format(self, fmt='YYYY-MM-DD HH:mm:ssZZ', locale='fr-ca'):
    return super().format(fmt, locale)
```

or using a function (this would be my preferred solution):

```python
def fr_ca_format(arw):
  return arw.format(locale='fr-ca')
```

eezomancer:
@Syeberman, hi! I made this PR in response to this [issue](https://github.com/crsmithdev/arrow/issues/309). If my understanding is correct, the goal is to make the locale persist beyond Arrow method calls so it can be used with python's String format.

A use case would be along the lines of:
```
>>> import arrow
>>> arw = arrow.get(locale='it_it')
>>> dict = {'name': 'Joe', 'today': arw}
>>> print("Hi, I'm {name}! Today is {today:dddd}.".format(**dict))
Hi, I'm Joe! Today is venerdì.
```

I suppose one could just change the `dict['today']` value to `arrow.get().format(fmt='dddd', locale='it_it')
` though. Wdyt?

eezomancer:
Closing so I can properly set up my local environment first. I'd still like to know if a PR for [this issue](https://github.com/crsmithdev/arrow/issues/309) is still worth pursuing.

PR #474, opened 2017-10-28
Added AlbanianLocaleTests

endri0x74:


systemcatch:
There is no Albanian locale implemented yet so these tests won't work. However if you felt like creating that locale it would be appreciated. That's going to require a new PR though.

PR #478, opened 2017-11-06
Escape regex special characters.

vilhelm:
Escape regex special characters in format string.

I ran into an issue when trying to parse a string in the format "YYYY(MM)" and it no longer would match the generated regex.

While it is currently possible to match under the current behavior by pre-escaping yourself, e.g., "YYYY\(MM\)", this seems non-intuitive.

This is applied after the [...] escaping to maintain the current behavior.

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -133,9 +133,9 @@ class DateTimeParser(object):
         escaped_fmt = re.sub('S+', 'S', escaped_fmt)
         escaped_data = re.findall(self._ESCAPE_RE, fmt)
 
-        fmt_pattern = escaped_fmt
+        fmt_pattern = re.escape(escaped_fmt)
 
-        for m in self._FORMAT_RE.finditer(escaped_fmt):
+        for m in self._FORMAT_RE.finditer(fmt_pattern):
             token = m.group(0)
             try:
                 input_re = self._input_re_map[token]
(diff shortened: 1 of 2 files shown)


systemcatch:
I believe this change was already done in #513, sorry your PR was ignored for so long.

jaycle:
Sorry to comment on a closed issue but apparently I was relying on this behavior in 0.12 which was fixed/broken in 0.13 Is there any way to match on something like: `MMM D[, ]+YYYY` which would match:
"Jan 1, 2019", "Jan 1 2019", and "Jan 1,2019". Is there any way to embed regex in the format string now? If not could we extend the api to allow for it?

vilhelm:
> Sorry to comment on a closed issue but apparently I was relying on this behavior in 0.12 which was fixed/broken in 0.13 Is there any way to match on something like: `MMM D[, ]+YYYY` which would match:
> "Jan 1, 2019", "Jan 1 2019", and "Jan 1,2019". Is there any way to embed regex in the format string now? If not could we extend the api to allow for it?

I would direct your comment to #513 since that PR was ultimately the one accepted.  It is possible to specify a list of formats, so for your case, you could do arrow.get('...', ['MMM D, YYYY', 'MMM DYYYY', 'MMM D YYYY'])

jaycle:
Some say comment the PR and others say the issue itself. I opted for the issue because I was going to open a new one.

I think that the list can work in some cases but it does not in cases where you want to say accept any amount of whitespace between date and year. If it would be considered, I can open a PR that allows for opting-in to such functionality.

PR #481, opened 2017-11-09
add day_of_year in parser

aysonje:
Resolves https://github.com/crsmithdev/arrow/issues/396

### DIFF
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -9,7 +9,7 @@ from arrow import util, locales
 
 class DateTimeFormatter(object):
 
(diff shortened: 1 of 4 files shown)


systemcatch:
Hey, sorry this has taken so long to be looked at. If we can fix the conflicts with master then I'd like to try and merge this in with a few changes. I'm not keen on having both `DDD` and `j` doing the same thing, better to just stick with `DDD` for consistency imo, %j can just be kept for working with the strptime class method.

We could also look at adding support for `DDDD` (zero padded version for day of year) but that can always be another PR.

systemcatch:
Parsing for the DDD and DDDD tokens has been implemented in the 0.15.0 branch.

PR #482, opened 2017-11-09
raise ParserError if year, month, or date is not found in string

armando1793:
this attempts to resolve https://github.com/crsmithdev/arrow/issues/456

using `arrow.get('31.12.2017')` as an example, the input string matches the format YYYY generated by parser.parse_isoformat. As a result, only year is parsed from the input string which leads to month and date defaulting to 1 in  _build_datetime.

This attempts to resolve the attached issue by raising parser error if year, date or month is missing from the  parts parameter provided to _build_datetime

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -180,7 +180,7 @@ class DateTimeParser(object):
             else:
                 value = match.group(token)
             self._parse_token(token, value, parts)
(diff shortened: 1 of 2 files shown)


armando1793:
closing. bad solution -_-

PR #485, opened 2017-11-09
include ZZZ in format example as a hint

smontanaro:
Extends the format example to use ZZZ so it's more readily apparent how to specify timezone names.

### DIFF
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -78,6 +78,11 @@ class ArrowFactory(object):
             >>> arrow.get('2013-09-29T01:26:43.830580')
             <Arrow [2013-09-29T01:26:43.830580+00:00]>
 
+        **One** ISO-8601-formatted ``str`` with tz name, to parse it::
+
+            >>> arrow.get('2013-11-29T01:26:43.830580 America/Chicago')
+            <Arrow [2013-11-29T01:26:43.830580-06:00]>
+
         **One** ``tzinfo``, to get the current time **converted** to that timezone::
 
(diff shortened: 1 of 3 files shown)


arrow/factory.py
```diff
@@ -78,6 +78,11 @@ def get(self, *args, **kwargs):
             >>> arrow.get('2013-09-29T01:26:43.830580')
             <Arrow [2013-09-29T01:26:43.830580+00:00]>
 
+        **One** ISO-8601-formatted ``str`` with tz name, to parse it::
+
+            >>> arrow.get('2013-11-29T01:26:43.830580 America/Chicago')
```
Syeberman: This is not an ISO-8601 compatible string, which states that timezones must be specified using offsets, not timezone names.  This `get` method is already quite magical, which works well but **only** because it relies on industry-standard inputs like timestamps and 8601.  I would be very wary of adding any additional magic to this function.

I'd be happy to accept a PR that just added your `arrow.get('2013-05-05 12:30:45 America/Chicago', 'YYYY-MM-DD HH:mm:ss ZZZ')` example to this file, but I'm not comfortable with the other changes here.

Additionally, if you could show that this string **is** indeed industry-standard, we could reopen this PR.

Syeberman: (Regardless, thanks for the PR!)

Syeberman:
I've created https://github.com/crsmithdev/arrow/pull/521 to merge the documentation tweak.  However, I'm rejecting the other changes.  (But thanks for the PR!)

PR #486, opened 2017-11-09
Drop datetime.utcnow in favor of datetime.now(UTC)

### LINKED ISSUE #477: Don't use utcnow
In the [code for `arrow.now()`](https://github.com/crsmithdev/arrow/blob/master/arrow/arrow.py#L82), I notice this:

```python
        utc = datetime.utcnow().replace(tzinfo=dateutil_tz.tzutc())
        dt = utc.astimezone(dateutil_tz.tzlocal() if tzinfo is None else tzinfo)
```

There's really no need for this, as you can just do this:

```python
tzinfo = tzinfo or dateutil_tz.tzlocal()
dt = datetime.now(dateutil_tz.tzlocal() if tzinfo is None else tzinfo)
```

pganssle:
This fixes #477. `datetime.utcnow()` is an abomination that returns a *naive* datetime representing the current time in UTC. It is very rarely what you want. Since `arrow` is returning aware datetimes from both of these operations, `datetime.now(the_desired_zone)` is preferable.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -79,8 +79,8 @@ class Arrow(object):
 
         '''
 
-        utc = datetime.utcnow().replace(tzinfo=dateutil_tz.tzutc())
-        dt = utc.astimezone(dateutil_tz.tzlocal() if tzinfo is None else tzinfo)
+        tzinfo = tzinfo or dateutil_tz.tzlocal()
+        dt = datetime.now(tzinfo)
 
         return cls(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
             dt.microsecond, dt.tzinfo)
@@ -92,10 +92,10 @@ class Arrow(object):
 
         '''
 
-        dt = datetime.utcnow()
+        dt = datetime.now(dateutil_tz.tzutc())
 
         return cls(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
-            dt.microsecond, dateutil_tz.tzutc())
+            dt.microsecond, dt.tzinfo)
 
     @classmethod
     def fromtimestamp(cls, timestamp, tzinfo=None):
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -62,7 +62,7 @@ class Arrow(object):
 
         if util.isstr(tzinfo):
             tzinfo = parser.TzinfoParser.parse(tzinfo)
-        tzinfo = tzinfo or dateutil_tz.tzutc()
+        tzinfo = tzinfo if tzinfo is not None else dateutil_tz.tzutc()
 
         self._datetime = datetime(year, month, day, hour, minute, second,
             microsecond, tzinfo)
@@ -79,7 +79,7 @@ class Arrow(object):
 
         '''
 
-        tzinfo = tzinfo or dateutil_tz.tzlocal()
+        tzinfo = tzinfo if tzinfo is not None else dateutil_tz.tzlocal()
         dt = datetime.now(tzinfo)
 
         return cls(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
@@ -112,7 +112,7 @@ class Arrow(object):
 
         '''
 
-        tzinfo = tzinfo or dateutil_tz.tzlocal()
+        tzinfo = tzinfo if tzinfo is not None else dateutil_tz.tzlocal()
         timestamp = cls._get_timestamp_from_input(timestamp)
         dt = datetime.fromtimestamp(timestamp, tzinfo)
 
@@ -151,7 +151,11 @@ class Arrow(object):
 
         '''
 
-        tzinfo = tzinfo or dt.tzinfo or dateutil_tz.tzutc()
+        if tzinfo is None:
+            if dt.tzinfo is None:
+                tzinfo = dateutil_tz.tzutc()
+            else:
+                tzinfo = dt.tzinfo
 
         return cls(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
             dt.microsecond, tzinfo)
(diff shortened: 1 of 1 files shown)


rhan-mentad:
On the other hand, more often than not we prefer naive datetime objects when we use the library in our personal use cases.

pganssle:
@rhan-mentad For one thing, that's not really relevant in this situation, since both of these functions return aware datetimes by default.

I also think it's a mismatch in abstractions to have a naive `datetime` that represents a time in a specific time zone. Naive datetime is useful for "free floating" times.

rhan-mentad:
According to pytz, ```The preferred way of dealing with times is to always work in UTC, converting to localtime only when generating output to be read by humans.```. I've also read some things on stack overflow about daylight savings conversion, speed of utc vs now which recommends using UTC.

Regarding the first set of changes it looks fine imo, I don't agree with removing utc in favor of now for the second one which implements utcnow though

pganssle:
@rhan-mentad That's not precisely true as a general rule, there are often times you want to work with timezone aware datetimes because they have the semantics you need, but that is not an issue here.

The issue is that the `utcnow()` function is a bit of a problem because it creates a timezone-aware datetime *as a naive datetime*. Presumably the reason for this is that until fairly recently, Python didn't ship any concrete timezone types, and so they wanted to provide an interface for "gentlemen's agreement" type `datetime` objects where it is implied but not explicitly specified that the datetime is in UTC.

There is no difference in the output, and it just confuses people to think that `utcnow()` is how you make a datetime in UTC. `utcnow` is equivalent to this:

```python
from datetime import datetime, timezone

def utcnow():
    dt = datetime.now(timezone.utc)
    return dt.replace(tzinfo=None)
```

What *arrow* is doing currently is thus equivalent to this:

```python
def arrow_utcnow():
    dt0 = datetime.now(timezone.utc)
    dt1 = dt0.replace(tzinfo=None)
    return dt1.replace(tzinfo=timezone.utc)  # The same thing as dt0
```

Using IPython's `%timeit`:

```python
In [7]: %timeit datetime.utcnow().replace(tzinfo=timezone.utc)
100000 loops, best of 3: 2.11 µs per loop

In [8]: %timeit datetime.now(timezone.utc)
1000000 loops, best of 3: 469 ns per loop
```

If other libraries are recommending you use `utcnow()`, they are wrong.

rhan-mentad:
Thanks for the clarification. Agree with the change if it is faster now.

arrow/arrow.py
```diff
@@ -79,8 +79,8 @@ def now(cls, tzinfo=None):
 
         '''
 
-        utc = datetime.utcnow().replace(tzinfo=dateutil_tz.tzutc())
-        dt = utc.astimezone(dateutil_tz.tzlocal() if tzinfo is None else tzinfo)
+        tzinfo = tzinfo or dateutil_tz.tzlocal()
```
Syeberman: This changes the behaviour of `tzinfo`.  Before, tzlocal was assumed if it was `None`, now tzlocal is used for anything falsy.

Can you change to this (or something equivalent):

```python
if tzinfo is None:
  tzinfo = dateutil_tz.tzlocal()
```

pganssle: Fair enough I suppose.

pganssle:
@Syeberman The change you requested (respecting falsy `tzinfo` objects) would have made that one fallback an outlier among all the other functions. I think it would be a somewhat misguided attempt at backward compatibility to keep the interface inconsistent just because it had previously been that way, so I've gone ahead and updated all of the instances I found in `arrow.py`. A more thorough scrubbing of the codebase may be warranted, I'm not sure.

arrow/arrow.py
```diff
@@ -151,7 +151,7 @@ def fromdatetime(cls, dt, tzinfo=None):
 
         '''
 
-        tzinfo = tzinfo or dt.tzinfo or dateutil_tz.tzutc()
+        tzinfo = tzinfo if tzinfo is not None else dt.tzinfo or dateutil_tz.tzutc()
```
Syeberman: I'm not certain what the order of operations is between `if` and `or`, and I expect most people won't either.  I feel this should have brackets around the `or`: `tzinfo if tzinfo is not None else (dt.tzinfo or dateutil_tz.tzutc())`.

pganssle: Actually, that would just have the same problem - if you are going to support falsy time-zones, the `dt.tzinfo` also need to be expanded into a `None` check.

pganssle: That said, disregarding falsy-`tzinfo`, the operator precedence doesn't actually matter in this case, because it's equivalent to a chained `or`:

1. `(tzinfo if tzinfo is not None else dt.tzinfo) or `dateutil_tz.tzutc()` 

2. `tzinfo if tzinfo is not None else (dt.tzinfo or `dateutil_tz.tzutc())` 

In both cases it will just use whichever the first one in the list is that is not `None`.

Syeberman: I'd be OK with that, as it ensures everything is explicit, but we know `dt.tzinfo` is either a `tzinfo` object or `None`, so it's not strictly necessary.

Unless we're concerned about someone creating a `tzinfo` subclass that can be falsy?  I took a look at the `datetime` docs and, while it states that `datetime`, `date`, and `time` are always truthy, it doesn't define that behaviour for `tzinfo`.  So, yeah, this might actually be required.

Syeberman: In `(tzinfo if tzinfo is not None else dt.tzinfo) or dateutil_tz.tzutc()`, if `tzinfo` is `0`, then `dateutil_tz.tzutc()` will be chosen.

With `tzinfo if tzinfo is not None else (dt.tzinfo or dateutil_tz.tzutc())`, if `tzinfo` is `0` then `dt.tzinfo` gets chosen.

pganssle: I said "disregarding falsy `tzinfo`". In the fixed version, passing `tzinfo=0` will be an error anyway. I was just saying that if you are operating under the assumption that `tzinfo` is an actual `datetime.tzinfo` (as the code has been up until now), the two are equivalent.

pganssle:
@Syeberman I suppose the non-lazy thing to do would be to add a test for this new behavior, but I don't know if I can get to that today, so if you prefer an earlier merge date, maybe we can save that for a separate PR.

Syeberman:
I'm comfortable merging as-is.  This "falsy" change is a pretty minor edge case IMO.

PR #496, opened 2017-11-25
Add functools-lru-cache to setup

ramonsaraiva:
https://github.com/crsmithdev/arrow/issues/495

This PR dynamically adds `'backports.functools_lru_cache==1.2.1'` to `install_requires` if python version is less than `3`.

This is urgent cause `arrow` is broken in `py2.x` versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 import codecs
 import os.path
 import re
+import sys
 
 try:
     from setuptools import setup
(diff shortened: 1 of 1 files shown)


FRidh:
Why is a specific patch release needed? Can't it work with any 1.2? Fixing to a patch release is bad practice. Also, why not allow the latest version?

PR #499, opened 2017-11-30
Relax dependency backports.functools-lru-cache to be >= 1.2.1

### LINKED ISSUE #497: backports.functools-lru-cache
Hello,,

Is there any way arrow could use backports.functools-lru-cache 1.4 (latest) for python 2.7

Carlos

rtdean:
`backports.functools-lru-cache` is up to version 1.4.  Because of how setuptools ends up calculating transitive dependencies, if you depend on something that depends on `backports.functools-lru-cache` (no version specified), and depend on something that depends on `arrow` that depends on `backports.functools-lru-cache==1.2.1`, you can end up with setuptools pulling in `backports.functools-lru-cache` at 1.4, and skipping/ignoring the `==1.2.1` requirement.

This then breaks execution, as package requirements aren't satisfied.

I've run the complete test suite on both linux and macos using 1.4 (latest), with no issue.  And, of course, the travis builds for this pr should pull in 1.4 as well, because of the relaxed requirements.

fixes #497

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ file_text = read(fpath('arrow/__init__.py'))
 
 install_requires = ['python-dateutil']
 if sys.version_info[0] < 3:
-    install_requires.append('backports.functools_lru_cache==1.2.1')
+    install_requires.append('backports.functools_lru_cache>=1.2.1')
 
 setup(
     name='arrow',
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ nose-cov==1.6
 chai==1.1.1
 sphinx==1.3.5
 simplejson==3.6.5
-backports.functools_lru_cache==1.2.1
+backports.functools_lru_cache>=1.2.1


setup.py
```diff
@@ -28,7 +28,7 @@ def grep(attrname):
 
 install_requires = ['python-dateutil']
 if sys.version_info[0] < 3:
-    install_requires.append('backports.functools_lru_cache==1.2.1')
+    install_requires.append('backports.functools_lru_cache==1.4')
```
xhochy: Would it be possible to have here `backports.functools_lru_cache>=1.2.1`?

rtdean: I personally don't have any issue with >= (in fact, I think less stringent bounding in general is a good thing, until a conflict is found), however all the other documented requirements are a strict ==.

I will make the change and see what the maintainer says.

MinchinWeb:
Note that declaring dependencies like this via runnable code means you can run into issue when generating wheels, specifically that the code to determine your dependencies will be run when the *when is generated* rather than when the wheel is installed.

Two solutions:
 - don't provide wheels
 - use the `extra_requires()` method as outlined [here](https://hynek.me/articles/conditional-python-dependencies/)

papachoco:
Any ETA for a release w/ this patch?

pganssle:
@papachoco Per @MinchinWeb, I would say this patch should not be required at all, and that #504 should be accepted instead. This is the wrong way to handle conditional dependencies.

ramonsaraiva:
Yes let's continue with https://github.com/crsmithdev/arrow/pull/504

PR #502, opened 2017-12-10
Makes the only_distance paramter value equivalent to value given to f…

### LINKED ISSUE #500: Using granularity in humanize causes it to ignore only_distance
Example:
```
>>> import arrow
>>> now = arrow.utcnow()
>>> then = now.shift(hours=-5)
>>> now.humanize(other=then, only_distance=True, granularity='hour')
'in 5 hours'
>>> now.humanize(other=then, only_distance=True)
'5 hours'
```
I expected both to return `5 hours`

Tested on 0.12.0 from pip in both python 2.7.12 and python 3.6.3.

nishantjain91:
The humanize function disregarded the value of only_distance when granularity value was also given. This request changed the default value("False") to match the given input value.

Also added  two testcases for the same.

This fixes #500 .

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -806,7 +806,7 @@ class Arrow(object):
             
             if(trunc(abs(delta)) != 1):
                 granularity += 's'
-            return locale.describe(granularity, delta, only_distance=False)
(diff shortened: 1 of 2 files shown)


Syeberman:
Thanks for the great PR @nishantjain91!

PR #508, opened 2018-01-02
update functools-lru-cache to avoid import problems fixes #497

### LINKED ISSUE #497: backports.functools-lru-cache

pablogamboa:
otherwise import fails after upgrading to 0.12

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ nose-cov==1.6
 chai==1.1.1
 sphinx==1.3.5
 simplejson==3.6.5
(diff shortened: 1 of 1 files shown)


pganssle:
This is a duplicate and suffers from the same problems as #499. #504 is a better way to fix this.

ramonsaraiva:
Yes, let's keep with the approach in #504 - thanks for the contribution anyway.

PR #511, opened 2018-01-18
get, now & utcnow will now use docs found in ArrowFactory

FHTMitchell:
Have taken the good documentation found in `arrow.factory.ArrowFactory` for the `get`, `now` and `utcnow` methods and replaced the somewhat lacking documentation for the wrappers in `arrow.api`. Now, IDEs and REPLs will produce the better documentation when asked to.

### DIFF
--- a/arrow/api.py
+++ b/arrow/api.py
@@ -8,12 +8,14 @@ methods for use as a module API.
 from __future__ import absolute_import
 
 from arrow.factory import ArrowFactory
+from functools import wraps
 
 
 # internal default factory.
 _factory = ArrowFactory()
 
 
+@wraps(_factory.get)
 def get(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


arrow/api.py
```diff
@@ -8,19 +8,23 @@
...
 
 from arrow.factory import ArrowFactory
+from functools import wraps
 
 
 # internal default factory.
 _factory = ArrowFactory()
 
 
+@wraps(_factory.get)
```
Syeberman: `@wraps` does more than just insert the documentation.  I'd prefer to see `get.__doc__ = _factory.get.__doc__`, to ensure that only the docstring is modified.

FHTMitchell: OK Agreed -- done

Syeberman:
Thanks @FHTMitchell!

PR #513, opened 2018-01-22
Fix RegEx parsing in the formatting string

sgnn7:
Fixes: https://github.com/crsmithdev/arrow/issues/501


Syeberman:
Thanks @sgnn7!  Note that I rebased and removed the version bump commit.  In general, only package maintainers bump the version number, and only during a release.

Cheers!

PR #518, opened 2018-02-21
Added Estonian localisation

zcribe:
Adapted version of hungarian locale.


Syeberman:
@zcribe, thanks for the patch!  The build is currently failing as there's some untested code in your submission.  Can you add a few more tests that target `_format_timeframe`?

Syeberman:
I believe, since all your `timeframes` are `dicts`, you just need to remove the `if isinstance(form,dict):` line.  Code coverage is likely telling you that the `isinstance` check is unnecessary.

Syeberman:
Thanks again, @zcribe!

PR #525, opened 2018-03-27
Fix spacing in deprecation warning

Callek:
I was fixing some deprecation warnings in code I use, and came across this formatting error.  (reads "valueis")


Syeberman:
Thanks @Callek!

PR #530, opened 2018-04-16
added latitude/longitude support for the creation and modification of…

czechben:
added to_geo and now_geo functions to the Arrow model which allows latitude/longitude specification as a way of describing timezone when creating a new arrow object. Also added a now_geo function to the factory which has same functionality. Created tests for each of the new methods in the respective test files.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -15,7 +15,6 @@ import calendar
 import sys
 import warnings
 
-import pytz
 from tzwhere import tzwhere
 
 from arrow import util, locales, parser, formatter
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -15,6 +15,8 @@ import calendar
 import sys
 import warnings
 
+import pytz
+from tzwhere import tzwhere
 
 from arrow import util, locales, parser, formatter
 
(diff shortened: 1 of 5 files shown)


pganssle:
Why did you add `pytz` here? I don't think Arrow even supports `pytz` zones. As a general rule I [caution people against pytz anyway](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html).

czechben:
removed pytz, not actually needed I forgot to remove the imports

pganssle:
I know arrow is an "everything and the kitchen sink" approach, but I personally prefer a cleaner API. It's not clear why you need to have the ability to get the time zone from latitude and longitude when you can just use `tzwhere` to get the time zone name from longitude and latitude.

If you're going to add this, you should probably make `tzwhere` an optional dependency, where it is either opportunistically used (e.g. it's not on the dependencies list but if it's installed you can use it) or you add it as a specific `extras` in the build, so that people can install `arrow[geo]` or `arrow[tzwhere]` and get the version that includes this feature (as well as the additional dependency).

Also, according to the CI it doesn't seem like `tzwhere` supports Python 2.6.

PR #531, opened 2018-04-16
Add Support for KWARGS construction (Response to issue#415)

ave233:
This pull request is a response to issue#415 (https://github.com/crsmithdev/arrow/issues/415). I have added support for construction of arrow objects using the same KWARGS format as the datetime object. Also added test cases to ensure the functionality of the changes along with meeting the QA expectations of 100% statement coverage.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -7,7 +7,7 @@ replacement.
 
 from __future__ import absolute_import
 
-from datetime import datetime, timedelta, tzinfo
+from datetime import datetime, timedelta, tzinfo, date
 from dateutil import tz as dateutil_tz
 from dateutil.relativedelta import relativedelta
 from math import trunc
@@ -191,6 +191,19 @@ class Arrow(object):
         return cls(dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second,
                    dt.microsecond, tzinfo)
 
+
+    @classmethod
+    def fromkwargs(cls, year=2013, month=5, day=5):
+        ''' Constructs an :class:`Arrow <arrow.arrow.Arrow>` object from a year string, month string,
+            and day string similar to the datetime constructor
+
+                :param year: the year.
+                :param month: the month.
+                :param day: the day.
+                '''
+
+        dt = date(year=year, month=month, day=day)
+        return cls(dt.year, dt.month, dt.day)
 
     # factories: ranges and spans
 
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -125,6 +125,9 @@ class ArrowFactory(object):
             >>> arrow.get(2013, 5, 5, 12, 30, 45)
             <Arrow [2013-05-05T12:30:45+00:00]>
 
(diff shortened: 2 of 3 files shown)


pganssle:
This is unnecessary. `Arrow` can already be constructed from kwargs:

```python
>>> from arrow import arrow
>>> Arrow(year=2014, month=12, day=1)
<Arrow [2014-02-01T00:00:00+00:00]>
```

I don't think you'd want to do it with a `fromkwargs` argument even if this weren't possible, though. And these are very weird and specific defaults...

ave233:
Arrow does not already support it. In fact it's worse then not supporting it. It still creates an arrow object but with the wrong properties instead of throwing an error.

I could be wrong here but, from the code snippet you just posted the Arrow object created had the incorrect information. It returned:
<Arrow [2014-02-01T00:00:00+00:00]>
When I believe it should have returned this:
<Arrow [2014-12-01T00:00:00+00:00]>


This is consistent with what the original issue report stated: "The current behaviour could introduce bugs because the kwargs do not raise an exception but return an object that is different from what the developer would expect." (https://github.com/crsmithdev/arrow/issues/415)

ave233:
@pganssle Thanks for the feedback. The ArrowFactory.get function is actually where I made the edits. I approached it a little differently but I can try changing it to use the Signature object like you suggested!

pganssle:
I'm mostly concerned that you added an unnecessary `fromkwargs` class method to `Arrow`, and that for whatever reason it uses May 5, 2013 as the default value...

ave233:
I thought I saw that as the default value in a few of the test cases but can obviously change that to anything you would like. I can also get rid of the fromkwargs class method and put the functionality in the ArrowFactory.get function

pganssle:
There should probably not be a default in this case, and yeah you don't need the `fromkwargs` because you can pass the keyword arguments directly to the `Arrow` class.

I think this is actually an easier bug to fix than you are making it.

The problem is [here](https://github.com/crsmithdev/arrow/blob/master/arrow/factory.py#L135). The argument count is showing up as 0 when you *only* pass keyword arguments. Look at this:

```python
>>> arrow.get(2014, 12, 4, minute=3)
<Arrow [2014-12-04T00:03:00+00:00]>
```

Basically what's happening is that the function signature is being inferred based on the number of positional arguments passed, ignoring the keyword arguments. Since none of the 0-, 1- and 2- argument constructors take keyword arguments, I think you can change it to:

```python
arg_count = len(args) + len(kwargs)
```

That will leave only one bug left, which is the case where 1 or 2 keyword arguments are passed. Not sure how the maintainers want to deal with this, but since 'year', 'month' and 'day' are all required, before the `if` statement, you can throw an exception in that case:

```python
if arg_count < 3 and len(kwargs):
    raise TypeError("Datetime-like constructor requires 'year', 'month' and 'day'")
```

ave233:
sounds good! I'll look into fixing it with that in mind

jadchaar:
Closing as this has been resolved in https://github.com/crsmithdev/arrow/pull/604.

Issue #498, opened 2017-11-29
span_range and interval return times outside the specified range


dwrpayne:
The implementation of span_range (and also interval()) is surprising to me, and is not what I would expect from a range function.

```
            >>> start = datetime(2013, 5, 5, 12, 30)
            >>> end = datetime(2013, 5, 5, 17, 15)
            >>> for r in arrow.Arrow.span_range('hour', start, end):
            ...     print(r)
            ...
            (<Arrow [2013-05-05T12:00:00+00:00]>, <Arrow [2013-05-05T12:59:59.999999+00:00]>)
            (<Arrow [2013-05-05T13:00:00+00:00]>, <Arrow [2013-05-05T13:59:59.999999+00:00]>)
            (<Arrow [2013-05-05T14:00:00+00:00]>, <Arrow [2013-05-05T14:59:59.999999+00:00]>)
            (<Arrow [2013-05-05T15:00:00+00:00]>, <Arrow [2013-05-05T15:59:59.999999+00:00]>)
            (<Arrow [2013-05-05T16:00:00+00:00]>, <Arrow [2013-05-05T16:59:59.999999+00:00]>)
            (<Arrow [2013-05-05T17:00:00+00:00]>, <Arrow [2013-05-05T17:59:59.999999+00:00]>)
```
I think the return value should be:
```
            (<Arrow [2013-05-05T12:30:00+00:00]>, <Arrow [2013-05-05T13:29:59.999999+00:00]>)
            (<Arrow [2013-05-05T13:30:00+00:00]>, <Arrow [2013-05-05T14:29:59.999999+00:00]>)
            (<Arrow [2013-05-05T14:30:00+00:00]>, <Arrow [2013-05-05T15:29:59.999999+00:00]>)
            (<Arrow [2013-05-05T15:30:00+00:00]>, <Arrow [2013-05-05T16:29:59.999999+00:00]>)
            (<Arrow [2013-05-05T16:30:00+00:00]>, <Arrow [2013-05-05T17:14:59.999999+00:00]>)
```

The first span should start at the specified `start`, and the final span should truncate so as not to extend beyond `end`.

I know "span" means something specific (`arrow.utcnow().span('day')`) and span_range() makes sense considering that definition, but this function still seems surprising and broken. Perhaps however I'm unaware of a use case for the existing span_range(), and I actually want to write a different function like split_range()?

At the very least, my use case is along the lines of:
```
start_date = arrow.get('2011-02-01')
[...]

PR #533, opened 2018-04-17
span_range() adds frame from start

Luca409:
As per issue #498 , under "I think the return value should be: ", span_range() should add the frame to the start time, rather than its current behavior. An optional "exact" flag has been added to the span_range() function, and this new behavior will occur when it is set to true.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ docs/_build/
 
 # VS Code
 .vscode/
+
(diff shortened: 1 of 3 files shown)


systemcatch:
@Luca409 
Sorry about the delay in looking at this. If you're still interested in getting this done can you fix the build and provide an example of the new output.

PR #544, opened 2018-09-09
Python 3.7 support.

krisfremen:
Hello,

This pull request adds Python 3.7 support.


sudoguy:
@krisfremen StopIteration is now replaced by RuntimeError (from python 3.7). But if you want to enable the same behaviour since python3.5 you can use 
```python
from __future__ import generator_stop
```

So, this exception won't be catch in python 3.7
https://github.com/crsmithdev/arrow/pull/544/files#diff-5d8eb9e97f8b648e4dae883cdce7b959R365

krisfremen:
hey guys, sorry. I moved countries in the past few weeks.

krisfremen:
@systemcatch ready to merge

systemcatch:
Thanks @krisfremen!

PR #550, opened 2018-11-03
Added Afrikaans locale

edwardvk:
Added a locale for the Afrikaans language in South Africa.


systemcatch:
Hey, thanks for the contribution @edwardvk. I don't have any experience or knowledge about Afrikaans but the code looks ok on a brief pass over.

Can you add some tests in https://github.com/crsmithdev/arrow/blob/master/tests/locales_tests.py

systemcatch:
I've updated this PR in #628, thanks for the contribution @edwardvk

PR #565, opened 2019-02-10
Moving macedonian locale to slavic base and adding tests

krisfremen:
Moving Macedonian to slavic base & adding tests


systemcatch:
Any particular reason for doing this? I'll wait until you're done making further changes to the new tests, 0.13.1 should be released shortly so this will be in the version after that.

krisfremen:
@systemcatch the translations are a bit awkwardly sounding when not using the slavic base.

Since it's a slavic language, it needs it to have the proper words for the different numbers.

systemcatch:
@krisfremen cool let me know when you're happy with this and I'll merge.

PR #571, opened 2019-02-22
Add additional timeframes translations for german language

niklasdoerfler:
When using the `humanize` method of arrow lib with the "only_distance" option set to true, in german language there are some special cases which have to be translated different than if not using the option.
For example "in one hour" in german is "in einer Stunde", with "only_distance" set to true, it is "one hour" which is translated to "einer Stunde" at the moment. But it would be better to translate it like "eine Stunde".
This PR adds these special cases to the german localisation.

### DIFF
--- a/arrow/locales.py
+++ b/arrow/locales.py
@@ -812,6 +812,13 @@ class _DeutschLocaleCommonMixin(object):
         'years': '{0} Jahren',
     }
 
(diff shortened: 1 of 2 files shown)


systemcatch:
Hey @niklasdoerfler thanks for the contribution! I'll merge this in shortly.

Issue #548, opened 2018-10-23
method to provide similar function as isBetween in moment.js


mubashirhanif:
I am working on some project, where I need to repeatedly check if a certain date is between 2 other dates.
Something like Moments.js's [isBetween](https://momentjs.com/docs/#/query/is-between/).

My current approach is something like, 
```
start = arrow.get(datetime.strptime(start_date, "%Y-%m-%d"))
end = arrow.get(datetime.strptime(end_date, "%Y-%m-%d"))
today = arrow.get(datetime.date.today())
r = arrow.Arrow.range('day', start, end)
return (today in r)
```

But I think this kind of functionality should come from arrow. I can make a pull request that introduces this method, let me know if it should be done.

PR #574, opened 2019-04-11
Add is_between functionality

jadchaar:
Addressing the feature request in #548, @TroyStacer and I have implemented the isbetween functionality in Arrow that functions similarly to the [isBetween function in Moment.js](https://momentjs.com/docs/#/query/is-between/).

Closes: #548


jadchaar:
Thanks for the feedback @systemcatch. I have addressed it in the latest commit.

Issue #541, opened 2018-08-03
humanize has odd behavior with future days


coopie:
example:
```python3
import arrow

print(arrow.utcnow().shift(days=1).humanize())
print(arrow.utcnow().shift(days=2).humanize())
print(arrow.utcnow().shift(days=3).humanize())
print(arrow.utcnow().shift(days=3, seconds=1).humanize())
print(arrow.utcnow().shift(days=4).humanize())
```

```
in a day
in 2 days
in 2 days
in 3 days
in 3 days
```

it seems the humanizer trips over after 2 days - after that it's one day behind every time.

looking into the code, the second delta is always one off being the right number of days.

for example the seconds delta between now and +3days is `259199`, making the delta `259200` seems to fix the problem

systemcatch:
The regression test for this issue very occasionally fails (see https://github.com/crsmithdev/arrow/pull/799/checks?check_run_id=742912216) which is annoying as it would break installs. Seems to happen only on Windows but no idea why yet.

/cc @jadchaar @krisfremen

PR #575, opened 2019-04-13
Fix humanize behavior with future days

TroyStacer:
After doing a little bit of debugging I found that the floating point delta value was being truncated by a cast to an int. If we simply round the number before casting this fixed the issue mentioned in https://github.com/crsmithdev/arrow/issues/541.

I have also included the test cases described in the issue thread.

Closes: #541

### DIFF
--- a/tests/arrow_tests.py
+++ b/tests/arrow_tests.py
@@ -1244,6 +1244,7 @@ class ArrowHumanizeTests(Chai):
         assertEqual(self.now.humanize(later, only_distance=True), '2 days')
         assertEqual(later.humanize(self.now, only_distance=True), '2 days')
 
+        # Regression tests for humanize bug referenced in issue 541
         later = self.now.shift(days=3)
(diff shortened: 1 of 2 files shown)


tests/arrow_tests.py
```diff
@@ -1244,6 +1244,15 @@ def test_days(self):
         assertEqual(self.now.humanize(later, only_distance=True), '2 days')
         assertEqual(later.humanize(self.now, only_distance=True), '2 days')
 
+        later = self.now.shift(days=3)
+        assertEqual(later.humanize(), 'in 3 days')
+
```
systemcatch: Maybe add a comment why these tests are needed (reference this PR/issue), that way people in the future won't remove them as unnecessary.

TroyStacer: Hi @systemcatch, thanks for the input, I just added a comment in the most recent commit.

TroyStacer:
Added a comments to test cases.

PR #576, opened 2019-04-14
Improved humanize behavior for near-zero durations

jadchaar:
Since @TroyStacer and I were already working on fixing issue #541 in Humanize, we thought we would tackle bug #416 as well since the PR and issue for it has been stale since mid 2017.

Fixes: https://github.com/crsmithdev/arrow/issues/416

The original PR for this issue (https://github.com/crsmithdev/arrow/pull/417) did not seem to fully address the issue and had failing tests, so we decided to open a new PR with fresh changes and updates.

### DIFF
--- a/arrow/locales.py
+++ b/arrow/locales.py
@@ -242,7 +242,7 @@ class EnglishLocale(Locale):
         :param only_distance: return only distance eg: "11 seconds" without "in" or "ago" keywords
         '''
 
-        humanized = super().describe(timeframe, delta, only_distance)
+        humanized = super(EnglishLocale, self).describe(timeframe, delta, only_distance)
         if only_distance and timeframe == 'now':
             humanized = 'instantly'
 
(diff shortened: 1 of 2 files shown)


jadchaar:
Thanks for the approval @systemcatch. One question though: would you like us to remove the docstrings comments from the inherited methods in the locales? Since describe is an internal method, it may not make much sense to keep the docstring comments. Would it be better to replace them with an inline comment explaining the rationale for the design decision?

systemcatch:
hey @jadchaar, sure if you feel like it but no need to go to any great effort.

jadchaar:
If I remove the docstrings from the two locales that override the describe method (English and German) to clean things up and reduce duplication, will Sphinx automatically copy over the parent classes docstring? I was reading online on a StackOverflow post that this seems to be the case, but I wanted to confirm :).

systemcatch:
@jadchaar I think it will copy over (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_inherit_docstrings) but it's fine to leave things as they are.

jadchaar:
@systemcatch I agree. The duplication is a bit bothersome, but I think it is better for the sake of usability in something like VSCode, even if the inheritance of docs works fine with Sphinx. This seems good to go then :D.

PR #583, opened 2019-05-03
Added new query functions

jadchaar:
I have added five new query functions to complement the recent addition of `is_between`. I feel that these additions will better align Arrow with Moment's query function interface.

One thing I would like feedback on: I did not know what to call the function argument other than "target". Any ideas for a better variable name is appreciated. Thanks!

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,0 +1,9 @@
+import arrow
+from datetime import datetime
+
+original = arrow.get(datetime(2013, 5, 7, 12, 30, 36))
+earlier = arrow.get(datetime(2013, 5, 7, 12, 30, 30))
+later = arrow.get(datetime(2013, 5, 7, 12, 30, 42))
+print(original.is_same_or_after(earlier))
+print(original.is_same_or_after(original))
+print(original.is_same_or_after(later))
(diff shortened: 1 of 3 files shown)


test.py
```diff
@@ -0,0 +1,9 @@
+import arrow
```
jaapz: Looks like you forgot to remove a test file here!

jadchaar:
@systemcatch thanks for the feedback. Are you referring to the overrided operators in `arrow.py`?

If so, I thought these would be a nice abstraction over those operators and a more friendly interface for those that are coming from Moment.js.

If you still think this does not add much value, feel free to close this PR and let me know what I could work on next :D.

systemcatch:
Yes I was referring to those operators, though I now understand what you wanted to achieve with these changes.

Personally I think python's syntax for this kind of thing is already quite user friendly, the is_between() is useful though. Let's leave this out for now.

I will have a think on some good issues to tackle.

jadchaar:
Would it be possible to triage some issues that you think should be tackled for the next release? I'd love to help out!

systemcatch:
> Would it be possible to triage some issues that you think should be tackled for the next release? I'd love to help out!

Sent you an email!

PR #587, opened 2019-05-19
Added flake8 to enforce formatting and style requirements

jadchaar:
- Made package and test source code flake8 compliant.
- Formatted code using [black](https://github.com/python/black) to help aid this process (black is made by a core Python language developer and enforces `pep8` style). I also had to make a number of manual tweaks to ensure flake8 compliance.
- Sorted imports using [isort](https://github.com/timothycrosley/isort).

I think flake8 compliance is important since it is good for newcomers to see a cleanly formatted codebase and it will allow us to enforce a code style and formatting requirement going forward.

A few notes:
- I had to add `self` to all the calls to Chai's `assert` functions to prevent flake8 from complaining about an undefined name.
- All regex strings had to be prepended with an `r` to prevent flake8 from complaining about inappropriate escape characters.
- black automatically changed all the single quotes to double quotes since it conforms with the `pep8` standard.
- flake8 does not like bare except statements in `try...except` blocks, but since I did not want to mess with functionality, I just added an inline comment to ignore flake8 when these are used. Ideally, I'd like to avoid the use of `noqa` ignore comments, but I thought that the bare excepts were needed since exceptions other than `ValueError` can be raised in particular circumstances where bare except statements are used. Correct me if I am wrong here though :).
- I added `pip` caching and flake8 linting to the Travis CI build.


hugovk:
How about also running Black on the CI to ensure new PRs keep it up?

```
black --check --diff
```

```
  --check                         Don't write the files back, just return the
                                  status.  Return code 0 means nothing would
                                  change.  Return code 1 means some files
                                  would be reformatted.  Return code 123 means
                                  there was an internal error.
  --diff                          Don't write the files back, just output a
                                  diff for each file on stdout.
```

jadchaar:
Thanks for the feedback @hugovk, I am a big fan of that. @systemcatch what do you think about Hugo's suggestion? I really like the idea of using `isort` and `black` checks in Travis CI to enforce style, but the only thing that makes me hesitant about this is that other popular packages like `requests` and `pytest` don't seem to enforce code style in Travis builds.

jadchaar:
Another idea could be to publish a pull request template and put formatting checks as a suggested requirement before submitting the PR (we can check it again in the Travis build).

systemcatch:
Hey @jadchaar this is a big change (but a good change I think), so I'd like to get @crsmithdev to sign off that he's happy with it.

Couple of questions;

- What is the line length limit we are enforcing?
- It seems black won't work for <3.6?

We could set flake8 to only test the latest code changes rather than checking the whole repo for every build, see https://stackoverflow.com/a/53978717

Possibly do the same for isort as well?

hugovk:
> Couple of questions;
> * What is the line length limit we are enforcing?

I'd suggest to go along with Black's suggested 88-char line length, most projects I've seen using Black go with it:

https://github.com/python/black#line-length

> * It seems black won't work for <3.6?

Yeah, Black only runs on Python 3.6+, but it can reformat code for Python 2 and 3.

https://github.com/python/black#installation-and-usage

hugovk:
> but the only thing that makes me hesitant about this is that other popular packages like `requests` and `pytest` don't seem to enforce code style in Travis builds.

`pytest` (10m downloads/month) does: https://travis-ci.org/pytest-dev/pytest/jobs/535150322#L239

`requests` doesn't use Black yet, but checking a couple of others from https://github.com/python/black#used-by, these do:

 * `attrs` (12m downloads/month): https://attrs.visualstudio.com/attrs/_build/results?buildId=80 (Test Lint)
 * PyPI (aka Warehouse) (serves 2,330m downloads/month): https://travis-ci.org/pypa/warehouse/jobs/535034842#L825

jadchaar:
Hugo covered your questions quite well Chris :). I really like how pytest and warehouse handle the enforcement of black, flake8, and isort (through pre-commits). This goes hand in hand with Chris's suggestion, but the libraries Hugo listed also seem to run this pre-commit hook on Travis.

systemcatch:
> I'd suggest to go along with Black's suggested 88-char line length, most projects I've seen using Black go with it:
Agreed let's go with that limit.

systemcatch:
@jadchaar feel free to make any tweaks you see fit, then this can be merged.

jadchaar:
Sorry for all the followup commits, but I finally got the system all configured with the help of @asottile at the pre-commit project. This system now automatically checks the changed code on commit via a pre-commit hook (which is configured when the virtual environment is first built). If this pre-commit hook is not run (e.g. the first build was not run or --no-verify was passed into the commit flag), the Travis CI build will run the same checks.

We are checking and enforcing the following:
Screen Shot 2019-05-26 at 1 18 02 AM

The flake8 check contains two plugins: flake8-bugbear, flake8-import-order.

Mind re-reviewing the changes @systemcatch @hugovk?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - repo: https://github.com/python/black
+      rev: 19.3b0
+      hooks:
+          - id: black
+            args: [--safe, --quiet, --target-version=py27]
+    - repo: https://gitlab.com/pycqa/flake8
+      rev: 3.7.7
+      hooks:
+          - id: flake8
+            additional_dependencies: [flake8-bugbear, flake8-import-order]
```
asottile: I wouldn't bother with `flake8-import-order` if you've also got `isort`

.travis.yml
```diff
@@ -1,21 +1,30 @@
...
 language: python
 dist: xenial
-python:
-  - 2.7
-  - 3.5
-  - 3.6
-  - 3.7
-  - 3.8-dev
+stages:
+    - linting
```
asottile: separate stages can be useful in some cases, but may frustrate your contributors (_personally_ I think things should be lint-clean before testing, but some like to throw code at CI just to see what sticks 🤷‍♂)

.isort.cfg
```diff
@@ -0,0 +1,4 @@
+[settings]
+known_third_party = chai,dateutil,pytz,setuptools,simplejson
+known_first_party = arrow
```
asottile: you shouldn't need this line I think (?) `isort` should sort the same without it

.travis.yml
jadchaar: Interesting. So you are saying that it may frustrate people if their code passes all the tests but the linting does not allow them to move to the next stage of CI?

.isort.cfg
jadchaar: Just tried it with only the `known_third_party` line and it still worked the same, so I cleaned up all the other settings :). Thanks for the feedback.

.travis.yml
asottile: nah, more that some times people want to throw a change at CI to see whether it passes tests or not, and not care about style until later -- if you have a blocking lint stage it would prevent that use case

jadchaar: Fair point. I have removed stages and went back to simply naming the linting stage with `name: "Linting"`.

jadchaar:
Done @systemcatch :).

PR #591, opened 2019-05-25
Upgraded dependencies to latest version and updated sphinx configurations

jadchaar:
In order to ensure that we are testing `arrow` with the latest dependencies that will be installed with `setup.py` (e.g. during `pip install`), I have updated the `requirements.txt` with the latest versions of all packages.

Since Sphinx was a massive upgrade, I re-ran `sphinx-quickstart` and transferred over the relevant pieces from the old `conf.py` and `Makefile`.

It may be worth exploring a change of doc theme from F6 in the future since it does not seem to be maintained anymore.


systemcatch:
Hey @jadchaar, I assume you've tried rebuilding the docs from scratch to check everything is working ok?

jadchaar:
@systemcatch I did do some testing on my end by running `make docs` and checking to see if the output and functionality of the produced HTML was similar and it seems to be working great.

That said, mind cloning my branch and testing out your usual document generation workflow to ensure that it is working as expected?

Also something interesting to note is that it seems like search has been improved in the new version of Sphinx:

Screen Shot 2019-05-26 at 12 43 57 PM
Screen Shot 2019-05-26 at 12 44 02 PM

Also Chris, it seems that https://arrow.readthedocs.io/en/latest/ is outdated and still on the 0.13.0 version of the docs. Any reason why that is the case?

jadchaar:
I made a few changes to `setup.py` that I forgot to make in my last PR. I thought they were not large enough to warrant a separate PR, so I added them onto this one. Added keywords and extra project URLs to point to on the PyPI page.

systemcatch:
Hey @jadchaar I built the docs from your branch and everything looks fine.

It seems that Sphinx has been dropped from the requirements.txt due to 2.0.1 no longer supporting Python 2.7, however that means people won't be able to make docs changes without manually installing it. We could try doing something like https://stackoverflow.com/a/33451105?

> Also Chris, it seems that https://arrow.readthedocs.io/en/latest/ is outdated and still on the 0.13.0 version of the docs. Any reason why that is the case?

readthedocs updates currently have to be manually pushed by @crsmithdev, I think automating this into every new release would be worthwhile if possible.

jadchaar:
@systemcatch I used the environment markers that you recommended in the latest commit. I originally removed `sphinx` from `requirements.txt` because there was no environment activation before the call to `make html`, so I did not think the dependency was needed since the tests did not require it. But you make a good point and I added the requirements back along with a call to the activation in the `Makefile` :).

systemcatch:
@jadchaar any preference on which of your PR's I merge first? I think if I do this one the conflicts should be easier for you to fix?

PR #597, opened 2019-06-03
Updated tox file and changed Travis CI to use tox

jadchaar:
Along with these tox configuration changes, I also updated the `isort` config to not conflict with Black (which wraps import lines beyond 88 characters and uses trailing commas).

Closes: https://github.com/crsmithdev/arrow/issues/589

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ matrix:
                   - $HOME/.cache/pre-commit
     allow_failures:
         - python: 3.8-dev
+          env: TOXENV=py38
 install: pip install -U codecov tox
-script: tox
+script: tox --recreate
 after_success: codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,5 @@ matrix:
         - python: 3.8-dev
           env: TOXENV=py38
 install: pip install -U codecov tox
-script: tox --recreate
+script: tox
 after_success: codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,24 +3,23 @@ dist: xenial
 matrix:
     include:
         - python: 2.7
+          env: TOXENV=py27
         - python: 3.5
+          env: TOXENV=py35
         - python: 3.6
+          env: TOXENV=py36
         - python: 3.7
+          env: TOXENV=py37
         - python: 3.8-dev
+          env: TOXENV=py38
         - name: "Linting"
           python: 3.7
-          env: TARGET=lint-ci
+          env: TOXENV=lint
           cache:
               directories:
                   - $HOME/.cache/pre-commit
     allow_failures:
         - python: 3.8-dev
-before_install: pip install -U codecov
-install:
-    - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then make build27; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.5' ]]; then make build35; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then make build36; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.7' ]]; then make build37; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.8-dev' ]]; then make build38; fi
-script: make "${TARGET:-test}"
+install: pip install -U codecov tox
+script: tox --recreate
 after_success: codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,5 @@ matrix:
     allow_failures:
         - python: 3.8-dev
 install: pip install -U codecov tox
-script: tox --recreate
+script: tox
 after_success: codecov
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,13 +8,11 @@ cover-package = arrow
 cover-erase = true
 
 [coverage:run]
-branch = True
-source =
-    arrow
-    tests
+branch = true
+source = arrow
 
 [coverage:report]
-show_missing = True
+show_missing = true
 fail_under = 100
 
 [flake8]
@@ -22,6 +20,9 @@ per-file-ignores = arrow/__init__.py:F401
 ignore = E203,E501,W503
 
 [tool:isort]
+line_length = 88
+multi_line_output = 3
+include_trailing_comma = true
 known_third_party = chai,dateutil,pytz,setuptools,simplejson
 
 [bdist_wheel]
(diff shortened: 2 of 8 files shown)


jadchaar:
@asottile I took your advice and decided to tweak our tox environment and use it for our Travis builds, but I am not sure why the builds are failing. Am I using tox incorrectly on Travis?

asottile:
the failure actually points out an important point of the packaging (and one of the many motivations for using `tox` in the first place!): **the packaging for arrow is currently broken when installing from source**

this line in the `setup.py`: https://github.com/crsmithdev/arrow/blob/cfafc0e4b2f2fcf2b004cb0ab2a67e2bb5feb56d/setup.py#L4

that attempts to import `arrow/__init__.py`

which in turn imports most of the arrow packaging

You generally have ~3 approaches here:

1. Read the file in `setup.py` instead of importing (PRO: no dependencies, CON: slightly janky? ish?)

    and you actually have two sub-approaches here, you can either use the ast or you can just read it and use a regex, the regex usually works fine 🤷‍♂ 

    ```python
    version_re = re.compile('(?<=__version__ = ")([^"]+)(?=")')
    with open('arrow/__init__.py') as f:
        version = version_re.search(f.read()).group()
    ```

   here's a way with an ast parser

    ```python
    class Visitor(ast.NodeVisitor):
        def visit_Assign(self, node):
            if (
                    len(node.targets) == 1 and
                    isinstance(node.targets[0], ast.Name) and
                    node.targets[0].id == '__version__' and
                    isinstance(node.value, ast.Str)
            ):
                self.version = node.value.s

    with open('arrow/__init__.py', 'rb') as f:
        visitor = Visitor()
        visitor.visit(ast.parse(f.read(), filename='arrow/__init__.py'))
        version = visitor.version
    ```

    it's more code, but it's more likely to not break due to regex fragility

2. only define the version in setup.py, and at runtime use `importlib_metadata.version(__name__)` to retrieve the version

    (this is the way I've used in my projects with success, though it does come with a slight import-time cost)  -- for example: https://github.com/pre-commit/pre-commit/blob/a75fe69984b1e1508cb89c41139bb8f384c86a53/pre_commit/constants.py#L21

3. Use setuptools-scm (or similar tool)

I don't know enough about how to set this up, but it's apparently a reasonable option

____

To demonstrate that it's broken from source:

```console
$ pip install arrow --no-binary :all:
Collecting arrow
  Downloading https://files.pythonhosted.org/packages/a0/3b/0b5c7771a619ee4eae1f894669dd5f7c0aae9c92d87891c4c746937f9cc4/arrow-0.14.0.tar.gz (102kB)
     |████████████████████████████████| 112kB 2.4MB/s 
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-pki32a8v/arrow/setup.py", line 4, in <module>
        from arrow import __version__
      File "/tmp/pip-install-pki32a8v/arrow/arrow/__init__.py", line 3, in <module>
        from .api import get, now, utcnow
      File "/tmp/pip-install-pki32a8v/arrow/arrow/api.py", line 10, in <module>
        from arrow.factory import ArrowFactory
      File "/tmp/pip-install-pki32a8v/arrow/arrow/factory.py", line 15, in <module>
        from dateutil import tz as dateutil_tz
    ModuleNotFoundError: No module named 'dateutil'
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-pki32a8v/arrow/
```

___

cheers! hope this helps :)

systemcatch:
@jadchaar does this need to go in 0.14.1 as well or can it wait until the next release?

jadchaar:
We can wait until the next release since this is more of a backend/housekeeping thing.

.travis.yml
```diff
@@ -3,24 +3,24 @@ dist: xenial
...
-install:
-    - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then make build27; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.5' ]]; then make build35; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then make build36; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.7' ]]; then make build37; fi
-    - if [[ $TRAVIS_PYTHON_VERSION == '3.8-dev' ]]; then make build38; fi
-script: make "${TARGET:-test}"
+          env: TOXENV=py38
+install: pip install -U codecov tox
+script: tox --recreate
```
asottile: I don't think you need `--recreate` here, I'm actually not sure why pytest does this 🤔

setup.cfg
```diff
@@ -8,20 +8,21 @@ cover-package = arrow
 cover-erase = true
 
 [coverage:run]
-branch = True
-source =
-    arrow
-    tests
```
asottile: I would strongly suggest instrumenting coverage for your tests -- it helps you find bugs that accidentally turn off portions of your testsuite

.travis.yml
jadchaar: I was wondering that as well. I thought it was unnecessary since each Travis run executes in a fresh VM.

jadchaar: Also, what do you think about adding the `TOXENV=py38` to `allow_failures` and `Linting`? I was unsure if I needed it in both.

setup.cfg
jadchaar: When I removed tests from the coverage run, the lines covered/hit did not change. Is it being instrumented incorrectly?

.travis.yml
asottile: looks like it's been there since the beginning of time in the pytest repo -- you don't need it (I'll go remove it from there too!)

asottile: your `allow_failures` looks correct to me :+1:

jadchaar: Oh sorry, I meant repeating `TOXENV=py38` in both `allow_failures` and in the `3.8-dev` target. They are both testing the same environment, so I was wondering if I need to repeat it. Let me try.

asottile: yeah you need it in both

jadchaar: Interesting. I wonder why hah

asottile: travis is picky and requires the whole mapping to match

jadchaar:
@systemcatch mind reviewing these changes so we can get them merged in for release `0.14.2`? Andrew found another regression with `setup.py` and locales (https://github.com/crsmithdev/arrow/pull/601).

jadchaar:
Note: I replaced `codecs` with `io` in `setup.py` in the latest commit after some research (namely https://stackoverflow.com/questions/5250744/difference-between-open-and-codecs-open-in-python).

PR #599, opened 2019-06-03
Fixed regression from 0.14.0

jadchaar:
@systemcatch @crsmithdev we need to get this patch pushed ASAP since a recent change broke the install.

Fixes: https://github.com/crsmithdev/arrow/issues/598

This other PR I am working on should prevent such as issue from coming up in the future: https://github.com/crsmithdev/arrow/pull/597

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,9 @@
 ## History
 
+### 0.14.1
+
+- [FIX] Fixed "ImportError: No module named 'dateutil'" (#598).
+
 ### 0.14.0
 
 - [NEW] Added provisional support for Python 3.8.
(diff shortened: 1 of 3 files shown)


jadchaar:
Proof of fix:

1) Screen Shot 2019-06-03 at 1 29 19 AM

2) The updated `setup.py` plays well with Tox now in this PR: https://github.com/crsmithdev/arrow/pull/597 (Tox uses the packaged version of arrow to run tests).

asottile:
> Oh dear, I wonder why that breakage wasn't caught by the tests.

the good news is @jadchaar is setting up `tox` which _will_ catch packaging issues like this in the future 🎉 (#597)

PR #601, opened 2019-06-04
Always open README with utf-8 encoding

andrewmwilson:
This is another regression in 0.14.0

Fixes: https://github.com/crsmithdev/arrow/issues/600

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
+import codecs
 import re
 
 from setuptools import setup
 
-with open("README.rst", "r") as f:
+with codecs.open("README.rst", encoding="utf-8") as f:
     readme = f.read()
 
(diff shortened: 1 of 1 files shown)


jadchaar:
@andrewmwilson  Looks good. Any difference between `with codecs.open` and `with open("README.rst", "r", encoding="utf-8") as f:`?

andrewmwilson:
@jadchaar I don't think the `encoding` parameter exists in Python 2.

There is also `io.open` in Python 2.6+.

jadchaar:
You are very right! Also, mind adding `encoding='utf-8'` to the reading of `init.py`?

Then we can get this merged in.

andrewmwilson:
Done!

jadchaar:
Thanks for the contribution @andrewmwilson. Apologies for these regressions; hopefully this will be the last one :).

Issue #415, opened 2017-01-26
arrow.get() should support kwargs 


LaurensBosscher:
From the docs:

> Arrow objects can be instantiated directly too, with the same arguments as a datetime:

I misread this as Arrow supporting the same syntax as datetime does: 

```
>>> datetime.datetime(year=2016, month=2, day=1)
datetime.datetime(2016, 2, 1, 0, 0)
```

However kwargs do not work as expected:

```
>>> import arrow
>>> arrow.get(year=2016, month=2, day=1)
<Arrow [2017-01-26T11:51:37.054519+00:00]>
```

Using kwargs is more readable (especially in international teams, is the 2 supposed to represent a month or day?) and would make refactoring code to use Arrow easier. 

The current behaviour could introduce bugs because the kwargs do not raise an exception but return an object that is different from what the developer would expect.

ave233:
Is there still interest in this issue? I'd be willing to work on this issue

PR #604, opened 2019-07-04
Allow factory.get() to construct with kwargs

systemcatch:
@jadchaar this is the small snippet of code I showed you a while back on zoom.

Essentially this lets us handle kwargs properly like the Arrow() constructor already does without having to do a ton of refactoring. Let me know if you think it's worth using as a quick fix.

```shell
arrow.get(year=2016, month=7, day=14)
<Arrow [2016-07-14T00:00:00+00:00]>
``` 

ref #415


systemcatch:
@jadchaar I removed the unnecessary pass and added extra test cases, let me know what you think.

PR #605, opened 2019-07-06
Replace theme with alabaster.

ashwineaso:
1. Replaced the documentation theme with alabaster


ashwineaso:
I was thinking of updating the docs to look similar to https://2.python-requests.org/en/master/. Please advice on what to include in this PR.

jadchaar:
The Alabaster theme is adapted from Kenneth Reitz's Requests theme, so it should be fine as is. Let me take a look at your PR.

docs/_themes/.gitignore
```diff
@@ -0,0 +1,3 @@
+*.pyc
```
jadchaar: This is fine, but these ignores are already included in the base directory's `.gitignore` file.

docs/conf.py
```diff
@@ -1,64 +1,250 @@
...
 
-html_theme = "f6"
-html_theme_path = ["_themes"]
+# Theme options are theme-specific and customize the look and feel of a theme
+# further.  For a list of options available for each theme, see the
+# documentation.
+#
+html_theme_options = {
+    "show_powered_by": False,
+    "github_user": "arrow",
```
jadchaar: This should probably be `crsmithdev`.

docs/index.rst
```diff
@@ -1,3 +1,8 @@
+.. arrow documentation master file, created by
```
jadchaar: Let's remove this comment to keep this file clean :).

docs/conf.py
ashwineaso: Yes. Sorry about that.

docs/index.rst
ashwineaso: Done.

ashwineaso:
Made all the changes you suggested. And cleaned up the mess, I hope. Please review once more.

jadchaar:
Looks good to me. Thanks for the contribution @ashwineaso 🎉!

PR #608, opened 2019-07-09
Added table of contents, added badges to docs, and general docs cleanup

jadchaar:
This is a followup PR to the recent conversion to the Alabster theme to add a couple of features and do some general cleanup.

Closes: https://github.com/crsmithdev/arrow/issues/593


jadchaar:
@systemcatch I changed the `__version__` string extraction method to match [what requests does](https://github.com/kennethreitz/requests/blob/master/setup.py#L60). It helps us to allow for changing the version only in one place, `_version.py`, rather than in `__init__.py` and in `docs/conf.py`. The issue we saw in `0.14.0` should not be present here since we are not importing `__init__.py`.

systemcatch:
Hey @jadchaar, I'm pretty busy until Thursday evening but should have a chance to look at this after then.

docs/index.rst
```diff
@@ -1,90 +1,17 @@
...
-    >>> local
-    <Arrow [2013-05-11T13:23:58.970460-07:00]>
-
-    >>> arrow.get('2013-05-11T21:23:58.970460+00:00')
-    <Arrow [2013-05-11T21:23:58.970460+00:00]>
+Release v\ |release|. (`Installation`_)
 
-    >>> local.timestamp
-    1368303838
+.. include:: ../README.rst
```
systemcatch: Ah good idea, reducing duplication between the readme and docs.

Issue #607, opened 2019-07-08
"a" token for am/pm is not being processed in get() when using Spanish locale


NestorTejero:
I have been parsing some dates in Spanish and found out that the [`a` token](https://arrow.readthedocs.io/en/latest/#tokens) for am/pm does not get applied with the locale `es`.

See for example this date in English. It gets parsed properly by doing 8 pm -> 20 h.

```
>>> arrow.get('June 30, 2019 - 08:00 pm', "MMMM DD, YYYY - hh:mm a")                                          
<Arrow [2019-06-30T20:00:00+00:00]>
```

However, in Spanish it does not work:

```
>>> arrow.get('Junio 30, 2019 - 08:00 pm', "MMMM DD, YYYY - hh:mm a", locale="es")                            
<Arrow [2019-06-30T08:00:00+00:00]>
```

In fact, anything can be put in the `a` token!

```
>>> arrow.get('Junio 30, 2019 - 08:00 pasdfasdfm', "MMMM DD, YYYY - hh:mm a", locale="es")                    
<Arrow [2019-06-30T08:00:00+00:00]>
```

systemcatch:
Hi @NestorTejero thanks for reporting this, I can reproduce it on the master branch.

From a quick glance at https://github.com/crsmithdev/arrow/blob/master/arrow/locales.py#L359 it looks like the Spanish locale is inheriting the `meridians` class variable from the base locale, meaning they are blank.

jadchaar:
@systemcatch would this be fixed by simply getting translations for `meridians = {"am": "", "pm": "", "AM": "", "PM": ""}`?

systemcatch:
@jadchaar yup that would fix it, I'll do a PR.

jadchaar:
This has been fixed in: https://github.com/crsmithdev/arrow/pull/610. This change should ship with `arrow 0.14.3`.

PR #610, opened 2019-07-12
Add meridians to Spanish locale

systemcatch:
```shell
arrow.get('Junio 30, 2019 - 08:00 pm', "MMMM DD, YYYY - hh:mm a", locale="es")
<Arrow [2019-06-30T20:00:00+00:00]>
```

/cc @NestorTejero
ref #607

### DIFF
--- a/arrow/locales.py
+++ b/arrow/locales.py
@@ -376,6 +376,8 @@ class SpanishLocale(Locale):
         "years": "{0} años",
     }
 
+    meridians = {"am": "am", "pm": "pm", "AM": "AM", "PM": "PM"}
+
     month_names = [
         "",
         "enero",
--- a/tests/parser_tests.py
+++ b/tests/parser_tests.py
@@ -727,6 +727,15 @@ class DateTimeParserMeridiansTests(Chai):
             parser_.parse("2013-01-01 5 DU", "YYYY-MM-DD h A"), datetime(2013, 1, 1, 17)
         )
 
+    # regression check for https://github.com/crsmithdev/arrow/issues/607
+    def test_es_meridians(self):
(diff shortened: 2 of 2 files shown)


jadchaar:
@systemcatch looks good, but I think it is worth adding a regression test (e.g. the example in the bug report) to ensure that this bug does not occur again in the future.

Also, do these changes fix the case of being able to add anything to the `a` token (e.g. `arrow.get('Junio 30, 2019 - 08:00 pasdfasdfm', "MMMM DD, YYYY - hh:mm a", locale="es")`)? Or is this something that we need to fix with the upcoming `get` fixes?

systemcatch:
The changes seem to fix that problem.
```shell
>>> arrow.get('Junio 30, 2019 - 08:00 pasdfasdfm', "MMMM DD, YYYY - hh:mm a", locale="es")
  File "<input>", line 1, in <module>
  File "/home/chris/arrow/arrow/api.py", line 21, in get
  File "/home/chris/arrow/arrow/factory.py", line 217, in get
    dt = parser.DateTimeParser(locale).parse(args[0], args[1])
  File "/home/chris/arrow/arrow/parser.py", line 191, in parse
    fmt_pattern_re.pattern, string
arrow.parser.ParserError: Failed to match '(?P<MMMM>(enero|febrero|marzo|abril|mayo|junio|julio|agosto|septiembre|octubre|noviembre|diciembre))\ (?P<DD>\d{2}),\ (?P<YYYY>\d{4})\ \-\ (?P<hh>\d{2}):(?P<mm>\d{2})\ (?P<a>(am|pm))' when parsing 'Junio 30, 2019 - 08:00 pasdfasdfm'
```

PR #614, opened 2019-07-14
Added escaping to docs

jadchaar:
I also removed the version locks from `requirements.txt` to force the latest versions of the packages to be installed.

Related to the discussion here: https://github.com/crsmithdev/arrow/issues/35


jadchaar:
I re-locked the dependencies @systemcatch :).

PR #616, opened 2019-07-16
Use update_wrapper instead of __doc__ setter

uburuntu:
Use more explicit `update_wrapper` instead of `get.__doc__ = _factory.get.__doc__`

Python 2: https://docs.python.org/2/library/functools.html#functools.update_wrapper
Python 3: https://docs.python.org/3/library/functools.html#functools.update_wrapper

### DIFF
--- a/arrow/api.py
+++ b/arrow/api.py
@@ -7,6 +7,8 @@ methods for use as a module API.
 
 from __future__ import absolute_import
 
+from functools import update_wrapper
+
 from arrow.factory import ArrowFactory
 
 # internal default factory.
@@ -21,7 +23,7 @@ def get(*args, **kwargs):
     return _factory.get(*args, **kwargs)
 
 
-get.__doc__ = _factory.get.__doc__
+update_wrapper(get, _factory.get)
 
 
 def utcnow():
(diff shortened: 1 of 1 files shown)


arrow/api.py
```diff
@@ -21,7 +23,7 @@ def get(*args, **kwargs):
     return _factory.get(*args, **kwargs)
 
 
-get.__doc__ = _factory.get.__doc__
+update_wrapper(get, _factory.get)
```
jadchaar: I was taking a look at the Python docs you linked, and it states that the `update_wrapper` function by default  "assigns to the wrapper function’s __name__, __module__ and __doc__, the documentation string) and WRAPPER_UPDATES (which updates the wrapper function’s __dict__, i.e. the instance dictionary)". Don't we just want the `__doc__` to be updated (e.g. `update_wrapper(get.__doc__, factory.get.__doc__)`)? Correct me if I am wrong though.

jadchaar:
Thanks for the contribution, but I have been thinking about this change and I think it is unnecessary because we only want to update the `__doc__` attribute, but `update_wrapper` copies way more than that according to the docs: https://docs.python.org/3/library/functools.html#functools.update_wrapper.

uburuntu:
It's ok :ok_hand:

PR #620, opened 2019-07-29
Move backports from extra_requires to install_requires

jadchaar:
This is now the official method for conditionally requiring python versions as documented here: https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies. Pytest does the same here: https://github.com/pytest-dev/pytest/blob/master/setup.py#L11.

Addresses: https://github.com/crsmithdev/arrow/issues/495

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
     ],
-    keywords="arrow date time datetime",
+    keywords="arrow date time datetime timestamp timezone",
     project_urls={
         "Repository": "https://github.com/crsmithdev/arrow",
         "Bug Reports": "https://github.com/crsmithdev/arrow/issues",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -23,8 +23,10 @@
     packages=["arrow"],
     zip_safe=False,
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*",
-    install_requires=["python-dateutil"],
-    extras_require={":python_version=='2.7'": ["backports.functools_lru_cache>=1.2.1"]},
+    install_requires=[
+        "python-dateutil",
+        "backports.functools_lru_cache;python_version=='2.7'",
```
systemcatch: Is it ok that we've removed the >=1.2.1 requirement here? Now it could install an older version.

jadchaar: I locked it to >=1.2.1 just in case :)

jadchaar:
Ready for re-review @systemcatch.

PR #621, opened 2019-07-29
Fix regression when tzinfo kwarg is a string

chyzzqo2:
In v 0.14.2 passing a date string, a format string and tzinfo kwarg where the value is a string worked correctly. This usage pattern broke in 0.14.3. This restores the previous behavior.


jadchaar:
Hey @chyzzqo2, thanks for the quick find and the fix. We will get this fix deployed in the next day or so.

jadchaar:
@chyzzqo2 `arrow 0.14.4` has been released to fix this issue. Thanks again for the PR.

PR #624, opened 2019-08-01
Removed deprecated replace shift functionality

jadchaar:
I am going through some of the issues mentioned in [this blog post from 2016](https://blog.eustace.io/please-stop-using-arrow.html) and I think it is time that we finally remove the ability to shift from the `replace` function in arrow. A deprecation warning has been in place since 2016, so I think it is an appropriate time to remove this functionality.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -9,7 +9,6 @@ from __future__ import absolute_import
 
 import calendar
 import sys
-import warnings
 from datetime import datetime, timedelta, tzinfo
 from math import trunc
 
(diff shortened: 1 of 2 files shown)


arrow/arrow.py
```diff
@@ -580,33 +579,17 @@ def replace(self, **kwargs):
         """
 
         absolute_kwargs = {}
-        relative_kwargs = {}  # TODO: DEPRECATED; remove in next release
```
systemcatch: 20 releases later it finally comes true!

damonlynch:
Thank you for continuing to improve Arrow. However, two points:

1. The documentation has not been updated. https://arrow.readthedocs.io/en/latest/#replace-shift
2. I had literally _no idea_ this removal was coming because I did not know it was even deprecated to begin with. I think you need to link prominently link to the change / history log from the documentation, and explicitly warn in documentation about breaking the API in an future release well before you break it!

jadchaar:
Hey @damonlynch, thanks for the feedback. A [warning has been set](https://github.com/crsmithdev/arrow/pull/624/files#diff-5d8eb9e97f8b648e4dae883cdce7b959L591) on the function since 2016 and it was well overdue for removal. We are trying to tackle the biggest complaints about `arrow` and this was one that was talked about quite a bit, so we thought it was time to finally remove it. I will submit a PR the clean up the docs and to improve the wording in the history file. We currently have the change noted [in our change log](https://github.com/crsmithdev/arrow/blob/master/HISTORY.md#0145), but the wording can be clarified.

PR #635, opened 2019-08-10
Updated HISTORY to reflect deprecation changes

jadchaar:
I added a bit more context to the history file to better describe the deprecation pushed in `v0.15.0`.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,8 +3,8 @@
 ### 0.14.5
 
 - [NEW] Added Afrikaans locale.
-- [CHANGE] Removed deprecated replace shift functionality.
-- [FIX] Fixed bug that occurred when factory.get() was passed a locale kwarg.
+- [CHANGE] Removed deprecated `replace` shift functionality. Users looking to pass plural properties to the `replace` function to shift values should use `shift` instead.
+- [FIX] Fixed bug that occurred when `factory.get()` was passed a locale kwarg.
 
 ### 0.14.4
 


catwell:
As a user, it wasn't clear that this was deprecated at all. [Your own documentation](https://arrow.readthedocs.io/en/latest) still uses plurals with replace in examples (and so those examples don't work with current release).

I also didn't expect a breaking change like this in a patch version. I do know the major is still 0, but still...

All that being said, I don't mean to rant, just to give a bit of feedback. I know you are lacking maintainers, and Arrow is a great project (which is why I use it), so thanks for everything!

jadchaar:
@catwell Thanks for the feedback. We included this in the patch version simply because the deprecation warning has been in place since 2016 and the code was supposed to be deprecated in an earlier major release. The old contributors and maintainers forgot about it, so we found the snippets of code and decided to finally axe them :). I agree though--we should have waited until 0.15.0. Apologies about that.

Do you think that you can point out some of the places in the documentation where the old replace functionality still exists? Any mentions of `replace` with plural parameters should be changed to `shift`.

catwell:
It looks like you fixed it in #637 but you could also remove this from the "features" section:

> Updated replace method with support for relative offsets, including weeks

I didn't get the depreciation warnings because my test framework wasn't showing them, I switched to pytest now so I should be aware next time.

jadchaar:
We have updated the documentation to reflect your feedback. Thanks again :).

PR #639, opened 2019-08-17
Updated regular expressions, fixed a few bugs, added tests for v0.15.0

jadchaar:
- `X` token can now properly parse float timestamp strings
- Time strings in basic and extended format are now parsed using a regular expression
- "Z" is no longer stripped from the end of the string to prevent mutating the passed in datetime string. I changed it to now split on the "Z" as well as the + and -, so the parsing logic now understands that a `tz` is present and will parse it accordingly (rather than going on the assumption that the logic defaults to UTC and ignores the Z, it is now more explicit with a `tzutc` object).
- Removed stripping of trailing whitespace
- Added a number of tests for regular expressions and the basic format
- `ParserError` exceptions that were thrown when the `day_of_year` logic encountered an error were not shown to the user because they were being prematurely caught. I fixed this by adding a new `ParserMatchError` that inherits from `ParserError` that can be caught by the internal logic and allow the `ParserErrors` to propagate.
- `ParserError` now inherits from `ValueError`, which makes more sense then a `RuntimeError`.
- Removed warnings and incremented version.


jadchaar:
Note: tests seem to be passing, but we are still not at 100% coverage.

arrow/parser.py
```diff
@@ -120,6 +122,7 @@ def parse_iso(self, datetime_string):
         has_tz = False
 
         # TODO: add tests for all the new formats, especially basic format
+        # NOTE: YYYYMM is omitted to avoid confusion with YYMMDD (no longer part of ISO 8601, but is still often used)
```
systemcatch: We don't test for `YYMMDD` in `parse_iso` though so `YYYYMM` could be allowed?

arrow/parser.py
```diff
@@ -141,53 +144,66 @@ def parse_iso(self, datetime_string):
...
 
             if has_space_divider:
                 date_string, time_string = datetime_string.split(" ", 1)
             else:
                 date_string, time_string = datetime_string.split("T", 1)
 
-            time_parts = re.split("[+-]", time_string, 1)
-            colon_count = time_parts[0].count(":")
+            time_parts = re.split(r"[\+\-Z]", time_string, 1, re.IGNORECASE)
+            # TODO: should we prevent mixing basic and extended formats? would need to ensure that dates, times, and timezones are in same format
```
systemcatch: I think we can leave that for a future release

arrow/parser.py
```diff
@@ -141,53 +144,66 @@ def parse_iso(self, datetime_string):
...
+            # use 'ZZ' token instead since tz offset is present in non-basic format
+            if len(time_parts) == 2 and ":" in time_parts[1]:
+                tz_format = "ZZ"
 
-            has_tz = len(time_parts) > 1
-            has_hours = len(time_parts[0]) == 2
-            has_minutes = colon_count == 1 or len(time_parts[0]) == 4
-            has_seconds = colon_count == 2 or len(time_parts[0]) == 6
-            has_subseconds = re.search("[.,]", time_parts[0])
+            time_components = self._TIME_RE.match(time_parts[0])
```
systemcatch: :clap: regex to the rescue again!

arrow/parser.py
```diff
@@ -351,33 +365,31 @@ def _build_datetime(parts):
 
         timestamp = parts.get("timestamp")
 
-        if timestamp:
+        if timestamp is not None:
```
systemcatch: any reason for changing this?

arrow/parser.py
```diff
@@ -429,23 +441,23 @@ def _choice_re(choices, flags=0):
 
 
 class TzinfoParser(object):
-
-    _TZINFO_RE = re.compile(r"([+\-])?(\d\d):?(\d\d)?")
+    # TODO: test against full timezone DB
```
systemcatch: that's a good idea

arrow/parser.py
```diff
@@ -429,23 +441,23 @@ def _choice_re(choices, flags=0):
...
+    _TZINFO_RE = re.compile(r"^([\+\-])?(\d{2})(?:\:?(\d{2}))?$")
```
systemcatch: again maybe we should go through this on zoom

docs/index.rst
```diff
@@ -374,6 +374,7 @@ Use the following tokens in parsing and formatting.  Note that they're not the s
 .. [#t2] localization support only for formatting
 .. [#t3] the result is truncated to microseconds, with `half-to-even rounding <https://en.wikipedia.org/wiki/IEEE_floating_point#Roundings_to_nearest>`_.
 .. [#t4] timezone names from `tz database <https://www.iana.org/time-zones>`_ provided via dateutil package
+.. [#t5] this token cannot be used for parsing timestamps out of natural language strings due to compatibility reasons
```
systemcatch: we should talk about this

arrow/parser.py
```diff
@@ -141,53 +144,66 @@ def parse_iso(self, datetime_string):
...
+                time_string = "HH{time_sep}mm{time_sep}ss{subseconds_sep}S".format(
+                    time_sep=time_sep, subseconds_sep=subseconds_sep
+                )
             elif has_seconds:
-                time_string = "HH:mm:ss"
+                time_string = "HH{time_sep}mm{time_sep}ss".format(time_sep=time_sep)
             elif has_minutes:
-                time_string = "HH:mm"
+                time_string = "HH{time_sep}mm".format(time_sep=time_sep)
             elif has_hours:
```
systemcatch: L191-194 is where coverage is failing.

arrow/parser.py
```diff
@@ -25,18 +34,20 @@ class DateTimeParser(object):
...
-    _TWO_DIGIT_RE = re.compile(r"\d{2}")
-    _TZ_RE = re.compile(r"[+\-]?\d{2}:?(\d{2})?|Z")
+    _TZ_Z_RE = re.compile(r"([\+\-])(\d{2})(?:(\d{2}))?|Z")
+    _TZ_ZZ_RE = re.compile(r"([\+\-])(\d{2})(?:\:(\d{2}))?|Z")
     _TZ_NAME_RE = re.compile(r"\w[\w+\-/]+")
+    # NOTE: timestamps cannot be parsed from natural language strings (by removing the ^...$) because it will
+    # break cases like "15 Jul 2000" and a format list (see issue #447)
     _TIMESTAMP_RE = re.compile(r"^\d+\.?\d+$")
-    # TODO: test timestamp thoroughly
+    _TIME_RE = re.compile(r"^(\d{2})(?:\:?(\d{2}))?(?:\:?(\d{2}))?(?:([\.\,])(\d+))?$")
```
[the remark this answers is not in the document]
jadchaar: For sure! As an appetizer, here is a `regex101` reference: https://regex101.com/r/kCaSBK/1/

arrow/parser.py
```diff
@@ -95,23 +106,14 @@ def __init__(self, locale="en_us", cache_size=0):
...
-        # strip leading and trailing whitespace
-        datetime_string = datetime_string.strip()
-
+        # TODO: add a flag to normalize whitespace (useful in logs, ref issue #421)
         has_space_divider = " " in datetime_string
         has_t_divider = "T" in datetime_string
 
         num_spaces = datetime_string.count(" ")
-        if has_space_divider and num_spaces != 1:
+        if has_space_divider and num_spaces != 1 or has_t_divider and num_spaces > 0:
```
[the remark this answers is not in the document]
jadchaar: Agreed, and I liked that at first! But after some thought, I was thinking that it may be better to keep the message broad ("Expected an ISO 8601-like string, but was given '____'. Try passing in a format string to resolve this.").

I'd be down for reverting this change, but I thought it would be worth a discussion.

arrow/parser.py
jadchaar: It could be, but I am worried about ambiguity. I think we should require a format string for ambiguous date formats that we are not 100% sure about. We don't want people complaining about `get` being erratic anymore :D.

arrow/parser.py
jadchaar: I changed it because `parts.get("timestamp")` will yield `None` if `timestamp` is not in the parts dictionary, and we only want to account for `None` cases here. It is best to be as explicit as possible :).

`if timestamp` handles more cases than just `None` cases (check out [this StackOverflow post](https://stackoverflow.com/questions/3901144/difference-between-if-x-and-if-x-is-not-none)).

arrow/parser.py
jadchaar: Yup! We can talk through this, but here is another appetizer: https://regex101.com/r/4NnaLU/1.

arrow/parser.py
jadchaar: L194 and L191->L194 are not reached, which means that L194 and the branch from L191->L194 are never triggered. Thinking about it, I have no idea how to even get it to L194. There was never an `else` condition before our new `0.15.0` changes, but we added it in. I don't think it can possibly be reached since all formatting issues should be caught by the `_TIME_RE` match above it (notice that I use the same error message since they technically are the same error of not matching time strings).

jadchaar: I changed the `elif has_hours` to an `else` case and removed the `ParserError`, similar to how it is in the [current `master` code base](https://github.com/crsmithdev/arrow/blob/master/arrow/parser.py#L108).

tests/parser_tests.py
```diff
@@ -832,54 +994,72 @@ def test_parse_iso_with_extra_words_at_start_and_end_invalid(self):
...
 
         self.assertEqual(
-            self.parser.parse_iso("20180517T105513Z"), datetime(2018, 5, 17, 10, 55, 13)
+            self.parser.parse_iso("20180517T10:55:13-07:00"),
+            datetime(2018, 5, 17, 10, 55, 13, tzinfo=tz.tzoffset(None, -25200)),
         )
 
         self.assertEqual(
-            self.parser.parse_iso("20180517T105513-0700"),
+            self.parser.parse_iso("20180517T105513-07:00"),
```
systemcatch: Basic format shouldn't have any colons I think.

tests/parser_tests.py
```diff
@@ -907,9 +1087,20 @@ def test_parse_utc(self):
...
         self.assertEqual(self.parser.parse("-01:00"), tz.tzoffset(None, -3600))
 
+        self.assertEqual(self.parser.parse("0100"), tz.tzoffset(None, 3600))
+        self.assertEqual(self.parser.parse("+0100"), tz.tzoffset(None, 3600))
+        self.assertEqual(self.parser.parse("-0100"), tz.tzoffset(None, -3600))
+
+        self.assertEqual(self.parser.parse("01"), tz.tzoffset(None, 3600))
+        self.assertEqual(self.parser.parse("+01"), tz.tzoffset(None, 3600))
+        self.assertEqual(self.parser.parse("-01"), tz.tzoffset(None, -3600))
+
```
systemcatch: 💪

arrow/parser.py
systemcatch: so for example someone may pass `"200806"` expecting `2020-08-06` but end up with `2008-06-01` ?

tests/parser_tests.py
jadchaar: oh yeah! I included 3 tests for mixing basic and extended formats and noted that it may raise an error in the future. I will just remove the 3 tests as they do not provide value to our suite.

arrow/parser.py
jadchaar: Yup, exactly!

Also, I am reading the ISO 8601 Wikipedia page and it says the following:  "By disallowing dates of the form YYYYMM, the standard avoids confusion with the truncated representation YYMMDD (still often used)."

It seems to be explicitly disallowed by the standard, so maybe I should change the comment to reflect that:
Screen Shot 2019-08-19 at 5 59 40 PM

systemcatch: Ah ok then fair enough let's avoid trying `YYYYMM`.

PR #643, opened 2019-08-19
Handle date with over 999999 microseconds

matthewdooler:
Fixes https://github.com/crsmithdev/arrow/issues/636

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -2,7 +2,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import re
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from dateutil import tz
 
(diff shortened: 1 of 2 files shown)


tests/parser_tests.py
```diff
@@ -315,6 +315,12 @@ def test_parse_subsecond_rounding(self):
         self.assertEqual(self.parser.parse(string, format), self.expected)
         self.assertEqual(self.parser.parse_iso(string), self.expected)
 
+        # overflow (zero out the subseconds and increment the seconds)
```
jadchaar: Let's add a comment to denote that this is a regression test for issue https://github.com/crsmithdev/arrow/issues/636 so it can be referenced in the future if the issue reappears.

arrow/parser.py
```diff
@@ -275,15 +275,27 @@ def _build_datetime(parts):
...
-        return datetime(
-            year=parts.get("year", 1),
-            month=parts.get("month", 1),
-            day=parts.get("day", 1),
-            hour=hour,
-            minute=parts.get("minute", 0),
-            second=parts.get("second", 0),
-            microsecond=parts.get("microsecond", 0),
-            tzinfo=parts.get("tzinfo"),
+        microsecond = parts.get("microsecond", 0)
```
jadchaar: Let's add a comment here to denote that we are accounting for the rounding up to `1000000` so that future contributors are aware of the new functionality :).

matthewdooler:
@jadchaar thanks for the suggestions, does this look good now?

jadchaar:
Looks great now!

Since this is a sensitive change, I wanna pass it by @systemcatch as well before merging it in.

PR #647, opened 2019-08-22
Updated docs, revamped contributing section, added doc linting

jadchaar:


docs/conf.py
```diff
@@ -14,7 +14,7 @@
 
 # -- Project information -----------------------------------------------------
 
-project = "arrow"
+project = "Arrow 🏹"
```
systemcatch: Is it safe to have an emoji in the conf file?

jadchaar: It breaks as it is now when the docs are built with Python 2.7, but I fixed it by changing it to `u"Arrow 🏹"` :). Docs build and display very nicely!

PR #649, opened 2019-08-24
Made a number of tweaks to the MANIFEST and tweaked a few error messages

jadchaar:
I updated the `MANIFEST.in` to follow a similar pattern to that of `requests`: https://github.com/psf/requests/blob/master/MANIFEST.in. I removed the docs because I do not see much value in wrapping them into the tarball.


systemcatch:
To be honest I think we need to leave the docs in the manifest since they've been in there from the start of the project.

PR #663, opened 2019-09-10
0.15.0 regression - Support negative numbers in the timestamp regular expression

JBKahn:
This solves https://github.com/crsmithdev/arrow/issues/662

Currently arrow does not allow you to pass in negative numbers for seconds relative to epoch when using strings. This patch is an attempt to fix this.

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -45,7 +45,7 @@ class DateTimeParser(object):
     _TZ_NAME_RE = re.compile(r"\w[\w+\-/]+")
     # NOTE: timestamps cannot be parsed from natural language strings (by removing the ^...$) because it will
     # break cases like "15 Jul 2000" and a format list (see issue #447)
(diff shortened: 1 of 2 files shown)


JBKahn:
@jadchaar if this is a-ok could you cut a release for this?

systemcatch:
Thanks @JBKahn, we'll look this over then merge if it's all good. We're probably going to wait a little before releasing 0.15.1 in case any more obvious bugs arise. Is this one really messing you guys up?

JBKahn:
Yeah, especially since a lot of apps don't lock down the version and we have lots of historical dates being passed into arrow.

systemcatch:
Alright we'll speed things along then.

JBKahn:
Thanks, already had some issues with some places automatically picking up 0.15.0 if they din't lock the version and it was a sub-dependancy.

PR #667, opened 2019-09-10
Add changelog to docs

### LINKED ISSUE #660: Add Changelog to arrow.readthedocs.io
To make it easier to find breaking changes between version...

systemcatch:
closes #660 

@jadchaar let me know if you're happy with the layout.

docs_with_changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 # Changelog
 
-## 0.15.1
+## 0.15.1 (2019-09-10)
 
 - [FIX] Fixed a bug that caused Arrow to fail when passed a negative timestamp string.
 - [FIX] Fixed a bug that caused Arrow to fail when passed a datetime object with `tzinfo` of type `StaticTzInfo`.
 
-## 0.15.0
+## 0.15.0 (2019-09-08)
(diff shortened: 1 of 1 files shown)


jadchaar:
This works, but I am thinking that we should embed it like `requests` does: https://2.python-requests.org//en/latest/community/updates/#software-updates.

On the [home page](https://2.python-requests.org/en/master/), `requests` has "Release history" under "Useful links". We could replicate that.

systemcatch:
Gotta say I think that works for requests because it's on a separate page, embedding into our only page might be too much. 

I like the way they include the release date alongside the version number, maybe we should do that?

jadchaar:
@systemcatch I agree that for now, the link is fine. In the future it my be worthwhile to explore another separate page in the docs for the changelog.

I think we should definitely add dates to it--at least for v0.14.0 and greater :).

systemcatch:
Yup ok I'll add dates to all the releases I can.

jadchaar:
Sounds good. Also I think we should add a `[BREAKING CHANGE]` tag like it was suggested in #660 to denote potentially big changes.

Issue #189, opened 2014-12-29
is it possible to load an isocalendar tuple into an arrow instance?


kurttheviking:
Currently, arrow conveniently provides:

```
>>> import arrow
>>> d = arrow.utcnow().isocalendar()
>>> d
(2015, 1, 1)
```

However, I am not sure if it is possible to load that tuple `d` later using `arrow.get()` &ndash; is there a format specification I am missing? For example:

```
>>> d_arrow = arrow.get(d, 'isocalendar')
```

If not, I am happy to help with a PR for a target spec.

andrewelkins:
Correct it currently can't handle a tuple, it expects a string. If you're still looking to make a PR, I'd start by looking at `arrow/factory.py`

kurttheviking:
ok great thank you; ill take a look

systemcatch:
https://stackoverflow.com/questions/304256/whats-the-best-way-to-find-the-inverse-of-datetime-isocalendar for some related discussion on this topic.

Simple approach is to convert the tuple to a date then pass it to fromdate().

jadchaar:
@systemcatch your PR #670 seems to have tackled this. We good to close this now?

systemcatch:
We're good, I'll close this.

PR #670, opened 2019-09-11
Allow factory.get() to create from a ISO calendar tuple

systemcatch:
ref #189 

**TODO**

- [x] Error handling.
- [x] Test edge cases.
- [x] Docstring.
- [x] Docs

### DIFF
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -199,7 +199,6 @@ class ArrowFactory(object):
 
             # (iso calendar) -> convert then from date
             elif isinstance(arg, tuple):
-                # TODO check len is 3 exactly or let function handle it?
                 dt = iso_to_gregorian(*arg)
                 return self.type.fromdate(dt)
 
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -179,11 +179,6 @@ class ArrowFactory(object):
             if isinstance(arg, date):
                 return self.type.fromdate(arg)
 
-            # # (iso calendar) -> convert then from date
-            # if isinstance(arg, tuple):
-            #     dt = iso_to_gregorian(*arg)
-            #     return self.type.fromdate(dt)
-
             # (tzinfo) -> now, @ tzinfo.
             elif isinstance(arg, dt_tzinfo):
                 return self.type.now(arg)
@@ -199,7 +194,7 @@ class ArrowFactory(object):
 
             # (iso calendar) -> convert then from date
             elif isinstance(arg, tuple):
-                # TODO check len is 3 exactly
+                # TODO check len is 3 exactly or let function handle it?
                 dt = iso_to_gregorian(*arg)
                 return self.type.fromdate(dt)
 
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -17,7 +17,7 @@ from dateutil import tz as dateutil_tz
 
 from arrow import parser
 from arrow.arrow import Arrow
-from arrow.util import is_timestamp, isstr
+from arrow.util import is_timestamp, iso_to_gregorian, isstr
 
 
 class ArrowFactory(object):
@@ -178,6 +178,10 @@ class ArrowFactory(object):
             # (date) -> from date.
             if isinstance(arg, date):
                 return self.type.fromdate(arg)
+
+            if isinstance(arg, tuple):
+                dt = iso_to_gregorian(*arg)
+                return self.type.fromdate(dt)
 
             # (tzinfo) -> now, @ tzinfo.
             elif isinstance(arg, dt_tzinfo):
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -179,9 +179,10 @@ class ArrowFactory(object):
             if isinstance(arg, date):
                 return self.type.fromdate(arg)
 
-            if isinstance(arg, tuple):
-                dt = iso_to_gregorian(*arg)
-                return self.type.fromdate(dt)
+            # # (iso calendar) -> convert then from date
+            # if isinstance(arg, tuple):
+            #     dt = iso_to_gregorian(*arg)
+            #     return self.type.fromdate(dt)
 
             # (tzinfo) -> now, @ tzinfo.
             elif isinstance(arg, dt_tzinfo):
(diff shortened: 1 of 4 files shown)


arrow/util.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
 
-from datetime import datetime
+import datetime
```
jadchaar: Any reason for this?

arrow/factory.py
```diff
@@ -99,6 +99,16 @@ def get(self, *args, **kwargs):
             >>> arrow.get(date(2013, 5, 5))
             <Arrow [2013-05-05T00:00:00+00:00]>
 
+        **One** time.struct time::
+
+            >>> arrow.get(gmtime(0))
+            <Arrow [1970-01-01T00:00:00+00:00]>
+
+        **One** iso calendar ``tuple``, to get that week date in UTC::
```
jadchaar: Let's capitalize ISO :)

arrow/factory.py
```diff
@@ -192,6 +197,11 @@ def get(self, *args, **kwargs):
             elif isinstance(arg, struct_time):
                 return self.type.utcfromtimestamp(calendar.timegm(arg))
 
+            # (iso calendar) -> convert then from date
+            elif isinstance(arg, tuple):
```
jadchaar: I am a bit worried that `tuple` is too broad. I wonder if there is any other datetime information we can absorb in the `get()` method that would be of type `tuple`.

arrow/util.py
```diff
@@ -12,12 +12,30 @@ def is_timestamp(value):
...
         return True
     except TypeError:
         return False
 
 
+# Credit to https://stackoverflow.com/a/1700069
+def iso_to_gregorian(iso_year, iso_week, iso_day):
+    """Converts an iso weekday tuple into a datetime object."""
+
+    if iso_week == 0 or iso_day == 0:
```
jadchaar: We should probably split this into two conditionals so we can make the error messages clearer.

tests/util_tests.py
```diff
@@ -29,3 +29,10 @@ def test_is_timestamp(self):
         overflow_timestamp_int = int(overflow_timestamp_float)
         with self.assertRaises((OverflowError, ValueError)):
             util.is_timestamp(overflow_timestamp_int)
+
+    def test_iso_gregorian(self):
```
jadchaar: We should probably test more edge cases such as those listed here: https://stackoverflow.com/q/304256/.

arrow/util.py
systemcatch: Mainly to make the imports easier.

systemcatch:
Hey @jadchaar I made most of the additions you suggested.

>I am a bit worried that `tuple` is too broad. I wonder if there is any other datetime information we can absorb in the `get()` method that would be of type `tuple`.

Looking through the `get()` method nothing else would be a `tuple`, `struct_time` is a subclass of `tuple` though so ordering does matter in the method. Maybe I could add a check like this to avoid any subclasses accidentally getting passed in the future.

```python
if issubclass(arg, tuple):
    # pass or raise here
```

arrow/util.py
```diff
@@ -12,12 +12,31 @@ def is_timestamp(value):
...
     except TypeError:
         return False
 
 
+# Credit to https://stackoverflow.com/a/1700069
+def iso_to_gregorian(iso_year, iso_week, iso_day):
+    """Converts an iso weekday tuple into a datetime object."""
+
+    if not 1 <= iso_week <= 53:
+        raise ValueError("Iso Calendar week value must be between 1-53.")
```
jadchaar: Let's make all the casings of ISO linear in the docstrings and error messages. We should probably capitalize it fully as ISO.

tests/factory_tests.py
```diff
@@ -140,6 +140,34 @@ def test_one_arg_iso_str(self):
 
         assertDtEqual(self.factory.get(dt.isoformat()), dt.replace(tzinfo=tz.tzutc()))
 
+    def test_one_arg_iso_calendar(self):
```
jadchaar: Solid tests, but we should also add the two mentioned in the Python docs: https://docs.python.org/3/library/datetime.html#datetime.date.isocalendar

arrow/util.py
```diff
@@ -12,12 +12,31 @@ def is_timestamp(value):
...
-        datetime.fromtimestamp(value)
+        datetime.datetime.fromtimestamp(value)
         return True
     except TypeError:
         return False
 
 
+# Credit to https://stackoverflow.com/a/1700069
+def iso_to_gregorian(iso_year, iso_week, iso_day):
+    """Converts an iso weekday tuple into a datetime object."""
```
jadchaar: Let's capitalize ISO in all the docstrings and error messages for consistency :)

arrow/factory.py
jadchaar: Seems like the best resolution to this is to expand this case to: `elif isinstance(arg, tuple) and len(tuple) == 3:` since ISO calendar is a 3-tuple.

jadchaar:
> Hey @jadchaar I made most of the additions you suggested.
> 
> > I am a bit worried that `tuple` is too broad. I wonder if there is any other datetime information we can absorb in the `get()` method that would be of type `tuple`.
> 
> Looking through the `get()` method nothing else would be a `tuple`, `struct_time` is a subclass of `tuple` though so ordering does matter in the method. Maybe I could add a check like this to avoid any subclasses accidentally getting passed in the future.
> 
> ```python
> if issubclass(arg, tuple):
>     # pass or raise here
> ```

I think since ISO calendar is a 3-tuple, it may be substantial enough to add a check for the length of the tuple like I showed in the comments above ^. If we add any tuples in the future, we can definitely investigate this further, but I think the length check is a good starting point.

systemcatch:
@jadchaar all changes now implemented!

arrow/factory.py
```diff
@@ -192,6 +197,11 @@ def get(self, *args, **kwargs):
             elif isinstance(arg, struct_time):
                 return self.type.utcfromtimestamp(calendar.timegm(arg))
 
+            # (iso calendar) -> convert then from date
```
jadchaar: Nit-picky detail, but what do you mean by "convert then"?

jadchaar: Never mind, I misread :).

PR #678, opened 2019-09-16
增加中文周，add chinese week

qingfai:


qingfai:
Added humanize week granularity translations for Chinese.

PR #685, opened 2019-10-02
Subsecond Parsing:

sidshrivastav:
- Added ability to parse subsecond token on basis of given number of 'S' string.
- Substitute of string with multiple 'S' to 'S' removed.
- While searching for pattern, to handle subsecond a condition is added to match the 'S+' pattern.
- For value, the value is being sliced by the length of token.

Closes: https://github.com/crsmithdev/arrow/issues/627


jadchaar:
@sidshrivastav hey thanks for the contribution. Do you think that you can add some tests to show this in action?

systemcatch:
Hey @sidshrivastav, I think we need a little more input from the community on this issue before proceeding with any new implementation. We'll review this at a later time.

Issue #594, opened 2019-05-26
Transition testing framework from nose/chai to pytest


jadchaar:
`nose` and `chai` are not regularly updated anymore, so a transition to [nose2](https://docs.nose2.io/en/latest/) or [pytest](https://docs.pytest.org/en/latest/) would be wise to explore.

For now, `nose` and `chai` seem to play fine with Python 2.7 through Python 3.8, but that could easily change in the future.

jadchaar:
It is worth noting that `nose` emits a couple of deprecation warnings in Python 3.7 that may become bigger issues once Python 3.8 is released.

systemcatch:
Yup just seen these warnings as well.
```shell
(arrow-37) chris@ThinkPad:~/arrow$ python -Wd -m nose
/home/chris/arrow-37/lib/python3.7/site-packages/nose/importer.py:12: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  from imp import find_module, load_module, acquire_lock, release_lock
/home/chris/arrow-37/lib/python3.7/site-packages/nose/suite.py:541: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  if isinstance(tests, collections.Callable) or isinstance(tests, unittest.TestSuite):
```
I'll look around to see if there's anyway to automatically convert nose/chai to another test suite.

jadchaar:
I have started to work on this transition and I ultimately decided on replacing Chai with Python Mock and nose with pytest.

We will see dependencies shift from `nose`, `nose-cov`, and `chai` to `pytest`, `pytest-cov`, and `pytest-mock`.

andrewelkins:
Need help with this transition? Was looking at this today when reviewing the requirements.txt for security vulnerabilities.

andrewelkins:
Besides pytest supporting most of nose out of the box, https://github.com/pytest-dev/nose2pytest might be of help.

jadchaar:
Hey @andrewelkins great to hear from you! Glad to see you back!

My feature branch is actually quite stale at this point since there have been tons of updates (including the big 0.15.0 release) after I began the transition.

[...]

PR #688, opened 2019-10-06
Formatter supports escape characters

### LINKED ISSUE #35: Allow escaping of characters in format()
In French, hours are formatted like "8 h 40" and not "8:40". But in an arrow's format string, _h_ display directly the hours. Is it possible to add a key like _!_ to force _h_ to be displayed normally ("hh !h mm")?

remeika:
The `DateTimeFormatter` now supports brackets as escape characters. Based almost entirely off the work of @bsidhom.

```python
>>> formatter = arrow.formatter.DateTimeFormatter()
>>> dt = datetime(2018, 3, 9, 8, 40)

>>> fmt = "YYYY-MM-DD h [h] m"
>>> formatter.format(dt, fmt)
u'2018-03-09 8 h 40'

>>> fmt = "YYYY-MM-DD h [hello] m"
>>> formatter.format(dt, fmt)
u'2018-03-09 8 hello 40'

>>> fmt = "YYYY-MM-DD h [hello world] m"
>>> formatter.format(dt, fmt)
u'2018-03-09 8 hello world 40'
```

One note: I could not get coverage to pass on of my lines of code, and had to add a no cover pragma. If anyone has an idea why code coverage is unhappy with [`formatter.py:31`](https://github.com/crsmithdev/arrow/pull/688/files#diff-98102a005603dcf5eb33f3152fe5099eR30), I'd love to fix it.

Fixes #35


systemcatch:
@remeika thanks for the PR, looks good from a quick glance over. Myself and Jad are pretty busy this week but hopefully can find some time to properly review it.

I wouldn't worry too much about the coverage problem since we're moving to pytest in #594 anyway.

arrow/formatter.py
```diff
@@ -12,8 +12,9 @@
 class DateTimeFormatter(object):
 
     _FORMAT_RE = re.compile(
-        r"(YYY?Y?|MM?M?M?|Do|DD?D?D?|d?dd?d?|HH?|hh?|mm?|ss?|SS?S?S?S?S?|ZZ?Z?|a|A|X)"
+        r"(\[(?:(?=(?P<literal>\\\\\[|\\\\\]|[^]]))(?P=literal))*\]|YYY?Y?|MM?M?M?|Do|DD?D?D?|d?dd?d?|HH?|hh?|mm?|ss?|SS?S?S?S?S?|ZZ?Z?|a|A|X)"
```
systemcatch: Can you walk us through this new pattern?

arrow/formatter.py
```diff
@@ -12,8 +12,9 @@
...
     )
+    _ESCAPED_BRACKET_RE = re.compile(r"\\\\(\[|\])")
```
systemcatch: Same for this one.

arrow/formatter.py
```diff
@@ -25,6 +26,11 @@ def format(cls, dt, fmt):
 
     def _format_token(self, dt, token):
 
+        if token and token.startswith("[") and token.endswith("]"):
+            return self._ESCAPED_BRACKET_RE.sub(
+                lambda m: m.group(1), token[1:-1]
```
systemcatch: Could you explain exactly what the lambda function is doing here? Maybe adding a few comments to the source code would help.

arrow/formatter.py
jadchaar: In particular, what is the purpose of the two pairs of escaped slashes?

jadchaar:
Hey @remeika, any update on this? We'd love to merge this in, but we have a few clarifying questions.

remeika:
@jadchaar Hey let me review the comments and reply later tonight!

remeika:
Sorry for the late response here @jadchaar @systemcatch ! I've removed all handling of escaped brackets, which I think was motivated by [this comment](https://github.com/crsmithdev/arrow/issues/35#issuecomment-25513900) by @bsidhom; as I said, this basic implementation was his. 

But I don't think handling escaped brackets is really necessary, so this feature now does not handle them at all, which really simplifies things. I think this code is now ready to go!

remeika:
@systemcatch As to why the last character of every escaped token is matching: I think this is an artifact of the workaround I'm using to emulate atomic capture groups. See the comment I added for details.

PR #695, opened 2019-10-21
Update formatter tests for tz abbreviations

### LINKED ISSUE #554: Investigate why timezone formatter sometimes fails
This is partly my fault as I missed rerunning travis before merging in #391 resulting in the build failing. I'll make sure to rerun tests on old PR's in the future.

The line that causes problems is https://github.com/crsmithdev/arrow/blob/master/tests/formatter_tests.py#L125

Something has clearly changed between now and when the tests first passed.

@rafaelsierra any ideas on this?

systemcatch:
`"BRST"` was never going to work in this test so I've substituted `"CST"` for now, this is a minor fix until we can test fully against the tz database.

closes #554

### DIFF
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -10,7 +10,7 @@ from arrow import locales, util
 
 
 class DateTimeFormatter(object):
-
+    # TODO: test against full timezone DB
     _FORMAT_RE = re.compile(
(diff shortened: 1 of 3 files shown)


jadchaar:
Curious, why is BRST never going to work?

Also, mind moving the TODO comment in the parser about testing against the full DB to the formatter so we remember?

systemcatch:
Moved the todo note.

Basically though BRT/BRST are used in day to day conversation the abbreviations aren't defined in the tz database (https://www.iana.org/time-zones) so they won't map correctly afaik. Brazil is also kinda crazy with all the timezone changes that have happened over the years.

PR #696, opened 2019-10-22
Added bounds parameter to Arrow.span_range, Arrow.interval and Arrow.span (#687)

### LINKED ISSUE #687: Closed / open intervals with span_range (feature request)
Hi,

Would it it be possible for ``span_range``, to return closed / open intervals instead of closed / closed interval (meaning the right part of the interval is open).

Could be achieved with an extra keyword.

For instance:

```
>>> for r in arrow.Arrow.span_range('hour', start, end, bounds='[)'):
...     print(r)
...
(<Arrow [2013-05-05T12:00:00+00:00]>, <Arrow [2013-05-05T13:00:00+00:00]>)
(<Arrow [2013-05-05T13:00:00+00:00]>, <Arrow [2013-05-05T14:00:00+00:00]>)
...
```

When dealing with intervals it is sometimes much more convenient to have closed / open ones.

Best regards,

Antoine

kkoziara:
The parameter allows to specify whether returned intervals should be closed, open, left-open or right-open. The parameter values and their semantics are identical to existing Arrow.in_between method.
Closes #687

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -351,7 +351,7 @@ class Arrow(object):
             )
 
     @classmethod
-    def span_range(cls, frame, start, end, tz=None, limit=None):
+    def span_range(cls, frame, start, end, tz=None, limit=None, bounds="[)"):
         """ Returns an iterator of tuples, each :class:`Arrow <arrow.arrow.Arrow>` objects,
         representing a series of timespans between two inputs.
 
@@ -361,6 +361,10 @@ class Arrow(object):
         :param tz: (optional) A :ref:`timezone expression <tz-expr>`.  Defaults to
             ``start``'s timezone, or UTC if ``start`` is naive.
         :param limit: (optional) A maximum number of tuples to return.
+        :param bounds: (optional) a ``str`` of either '()', '(]', '[)', or '[]' that specifies
+            whether to include or exclude the start and end values in the span. '(' excludes
+            the start, '[' includes the start, ')' excludes the end, and ']' includes the end.
+            If the bounds are not specified, the default bound '[)' is used.
 
         **NOTE**: The ``end`` or ``limit`` must be provided.  Call with ``end`` alone to
         return the entire range.  Call with ``limit`` alone to return a maximum # of results from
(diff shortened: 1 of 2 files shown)


jadchaar:
Hey @kkoziara, thanks for the contribution! @systemcatch and I have been super busy lately, but things should clear up next week and we can review it then :). Thanks!

kkoziara:
Cool :) I am glad to help.

jadchaar:
@kkoziara don't think that we forgot your two PRs :). We both did a first pass review and things seems solid, but we have just been very busy lately to give it a hard final look. Stay tuned!

tests/arrow_tests.py
```diff
@@ -1207,6 +1207,22 @@ def test_aware_tz(self):
             self.assertEqual(f.tzinfo, tz.gettz("US/Central"))
             self.assertEqual(c.tzinfo, tz.gettz("US/Central"))
 
+    def test_bounds_param_is_passed(self):
```
systemcatch: Hey @kkoziara, sorry for the delay. Quick question, the reason you're mainly testing on `span` is because both `span_range` and `interval` use `span` under the hood right?

kkoziara: Yes, exactly

arrow/arrow.py
```diff
@@ -411,6 +415,10 @@ def interval(cls, frame, start, end, interval=1, tz=None):
         :param end: (optional) A datetime expression, the end of the range.
         :param interval: (optional) Time interval for the given time frame.
         :param tz: (optional) A timezone expression.  Defaults to UTC.
+        :param bounds: (optional) a ``str`` of either '()', '(]', '[)', or '[]' that specifies
+            whether to include or exclude the start and end values in the span. '(' excludes
```
systemcatch: I think it needs to be interval instead of span in the docstring.

arrow/arrow.py
```diff
@@ -361,6 +361,10 @@ def span_range(cls, frame, start, end, tz=None, limit=None):
         :param tz: (optional) A :ref:`timezone expression <tz-expr>`.  Defaults to
             ``start``'s timezone, or UTC if ``start`` is naive.
         :param limit: (optional) A maximum number of tuples to return.
+        :param bounds: (optional) a ``str`` of either '()', '(]', '[)', or '[]' that specifies
+            whether to include or exclude the start and end values in the span. '(' excludes
```
systemcatch: I think it needs to be span range instead of span in the docstring.

arrow/arrow.py
kkoziara: Maybe "intervals" because it concerns every interval?

arrow/arrow.py
kkoziara: Similarly to above comment, maybe "in each span in the range" as it's about spans not the range.

arrow/arrow.py
systemcatch: Sounds good.

arrow/arrow.py
systemcatch: Again I like your phrasing.

Issue #657, opened 2019-09-09
Test TzinfoParser against full timezone database


systemcatch:
See https://github.com/crsmithdev/arrow/blob/master/arrow/parser.py#L454

Some ideas on how to do this (https://stackoverflow.com/questions/15453917/get-all-available-timezones), seems like both dateutil and pytz could be used. I tend towards using pytz because it's independent of the operating system and we already use it in other tests.

certator:
I will give it a try 👷

PR #699, opened 2019-10-25
Test formatter against full timezone database #693

### LINKED ISSUE #693: Test formatter against full timezone database
Similar to https://github.com/crsmithdev/arrow/issues/657, it would be great to test the timezone formatter against this DB as well.

Code for tests could go here: https://github.com/crsmithdev/arrow/blob/master/tests/formatter_tests.py#L122.

haneybarg:
Closes #693 
Similar to #657, zones are retrieved via dateutil and pytz, and are used in existing test. The abbreviation is also retrieved from the existing test, using the method `tzname()`.

### DIFF
--- a/tests/formatter_tests.py
+++ b/tests/formatter_tests.py
@@ -5,8 +5,15 @@ from datetime import datetime
 import pytz
 from chai import Chai
 from dateutil import tz as dateutil_tz
+from dateutil.zoneinfo import get_zonefile_instance
 
 from arrow import formatter
+
+
+def make_full_tz_list():
+    dateutil_zones = set(get_zonefile_instance().zones)
+    pytz_zones = set(pytz.all_timezones)
+    return dateutil_zones.union(pytz_zones)
 
 
 class DateTimeFormatterFormatTokenTests(Chai):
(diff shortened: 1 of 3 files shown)


tests/formatter_tests.py
```diff
@@ -5,10 +5,17 @@
 import pytz
 from chai import Chai
 from dateutil import tz as dateutil_tz
+from dateutil.zoneinfo import get_zonefile_instance
 
 from arrow import formatter
 
 
+def make_full_tz_list():
```
jadchaar: Let's import this from `parser_tests.py` so we do not repeat code. Or better yet, we could make a `utils.py` file and import from it in both `formatter_tests.py` and `parser_tests.py`. Either would work, but the second option may be more maintainable.

haneybarg: Hi! Ok, done.

tests/formatter_tests.py
```diff
@@ -7,6 +7,7 @@
 from dateutil import tz as dateutil_tz
 
 from arrow import formatter
+from tests.utils import make_full_tz_list
```
jadchaar: This works, but it may be better to say `from .utils import make_full_tz_list` since you are importing from the same directory.

haneybarg: Done!

PR #700, opened 2019-10-28
Add support for negative timestamps on Windows (#675)

### LINKED ISSUE #675: Add support for negative timestamps on Windows (before the epoch)
Currently, on Windows, you get the following error if you pass a negative timestamp:

```shell
(arrow) C:\Users\Chris>python
Python 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 20:34:20) [MSC v.1916 64 bit (AMD64)] on win32
>>> import arrow
>>> arrow.get(-15675657)
  File "C:\Users\Chris\arrow\lib\site-packages\arrow\api.py", line 21, in get
  File "C:\Users\Chris\arrow\lib\site-packages\arrow\factory.py", line 155, in get
  File "C:\Users\Chris\arrow\lib\site-packages\arrow\arrow.py", line 166, in utcfromtimestamp
OSError: [Errno 22] Invalid argument
```

This occurs on all versions of Arrow on Windows and seems to be due to limitations of the `fromtimestamp` and `utcfromtimestamp` methods in the datetime library on Windows.

This functionality works fine on Linux and MacOS:

```shell
venv ❯ python
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> import arrow
>>> import time
>>> arrow.get(time.time())
<Arrow [2019-09-14T15:41:17.997282+00:00]>
>>> arrow.get(time.time() * -1)
<Arrow [1920-04-19T08:18:37.826506+00:00]>
```

kkoziara:
Adds support for negative timestamps on Windows by computing the `datetime` directly using `timedelta`.  This is method is only used as a fallback, so on other platforms or with positive timestamps the `datetime.fromtimestamp` or `datetime.utcfromtimestamp` methods are used.
Closes #675

### DIFF
--- a/arrow/util.py
+++ b/arrow/util.py
@@ -2,6 +2,11 @@
 from __future__ import absolute_import
 
 import datetime
+import math
+import os
+import time
+
+import dateutil
 
 
 def total_seconds(td):  # pragma: no cover
@@ -21,6 +26,42 @@ def is_timestamp(value):
         return True
     except ValueError:
         return False
+
+
+def datetime_from_timestamp(timestamp, tz=None):
+    """Computes datetime from timestamp. Supports negative timestamps on Windows platform."""
+    sec_frac, sec = math.modf(timestamp)
+    dt = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()) + datetime.timedelta(
+        seconds=sec, microseconds=sec_frac * 1e6
+    )
+    if tz is None:
+        tz = dateutil.tz.tzlocal()
+    if tz == dateutil.tz.tzlocal():
+        # because datetime.astimezone does not work on Windows for tzlocal() and dates before the 1970-01-01
+        # take timestamp from appropriate time of the year, because of daylight saving time changes
+        ts = time.mktime(dt.replace(year=1970).timetuple())
+        dt += datetime.datetime.fromtimestamp(ts) - datetime.datetime.utcfromtimestamp(
+            ts
+        )
+        return dt.replace(tzinfo=dateutil.tz.tzlocal())
+    else:
+        return dt.astimezone(tz)
+
+
+def safe_utcfromtimestamp(timestamp, os_name=os.name):
+    """ datetime.utcfromtimestamp alternative which supports negatvie timestamps on Windows platform."""
+    if os_name == "nt" and timestamp < 0:
+        return datetime_from_timestamp(timestamp, dateutil.tz.tzutc())
+    else:
+        return datetime.datetime.utcfromtimestamp(timestamp)
+
+
+def safe_fromtimestamp(timestamp, tz=None, os_name=os.name):
+    """ datetime.fromtimestamp alternative which supports negatvie timestamps on Windows platform."""
+    if os_name == "nt" and timestamp < 0:
+        return datetime_from_timestamp(timestamp, tz)
+    else:
+        return datetime.datetime.fromtimestamp(timestamp, tz)
 
 
 # Credit to https://stackoverflow.com/a/1700069
@@ -57,4 +98,12 @@ except NameError:  # pragma: no cover
         return isinstance(s, str)
 
 
-__all__ = ["total_seconds", "is_timestamp", "isstr", "iso_to_gregorian"]
+__all__ = [
+    "total_seconds",
+    "is_timestamp",
+    "isstr",
+    "iso_to_gregorian",
+    "datetime_from_timestamp",
+    "safe_utcfromtimestamp",
+    "safe_fromtimestamp",
+]
--- a/arrow/util.py
+++ b/arrow/util.py
@@ -28,30 +28,30 @@ def is_timestamp(value):
         return False
 
 
-def datetime_from_timestamp(timestamp, tz=None):
+def windows_datetime_from_timestamp(timestamp, tz=None):
     """Computes datetime from timestamp. Supports negative timestamps on Windows platform."""
     sec_frac, sec = math.modf(timestamp)
     dt = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()) + datetime.timedelta(
         seconds=sec, microseconds=sec_frac * 1e6
     )
-    if tz is None:
-        tz = dateutil.tz.tzlocal()
-    if tz == dateutil.tz.tzlocal():
-        # because datetime.astimezone does not work on Windows for tzlocal() and dates before the 1970-01-01
-        # take timestamp from appropriate time of the year, because of daylight saving time changes
-        ts = time.mktime(dt.replace(year=1970).timetuple())
-        dt += datetime.datetime.fromtimestamp(ts) - datetime.datetime.utcfromtimestamp(
-            ts
-        )
-        return dt.replace(tzinfo=dateutil.tz.tzlocal())
-    else:
-        return dt.astimezone(tz)
+    if tz is not None:
+        if tz == dateutil.tz.tzlocal():
+            # because datetime.astimezone does not work on Windows for tzlocal() and dates before the 1970-01-01
+            # take timestamp from appropriate time of the year, because of daylight saving time changes
+            ts = time.mktime(dt.replace(year=1970).timetuple())
+            dt += datetime.datetime.fromtimestamp(
+                ts
+            ) - datetime.datetime.utcfromtimestamp(ts)
+            dt = dt.replace(tzinfo=dateutil.tz.tzlocal())
+        else:
+            dt = dt.astimezone(tz)
+    return dt
 
 
 def safe_utcfromtimestamp(timestamp, os_name=os.name):
     """ datetime.utcfromtimestamp alternative which supports negatvie timestamps on Windows platform."""
     if os_name == "nt" and timestamp < 0:
-        return datetime_from_timestamp(timestamp, dateutil.tz.tzutc())
+        return windows_datetime_from_timestamp(timestamp, dateutil.tz.tzutc())
     else:
         return datetime.datetime.utcfromtimestamp(timestamp)
 
@@ -59,7 +59,10 @@ def safe_utcfromtimestamp(timestamp, os_name=os.name):
 def safe_fromtimestamp(timestamp, tz=None, os_name=os.name):
     """ datetime.fromtimestamp alternative which supports negatvie timestamps on Windows platform."""
     if os_name == "nt" and timestamp < 0:
-        return datetime_from_timestamp(timestamp, tz)
+        if tz is None:
+            # because datetime.fromtimestamp default is local time
+            tz = dateutil.tz.tzlocal()
+        return windows_datetime_from_timestamp(timestamp, tz)
     else:
         return datetime.datetime.fromtimestamp(timestamp, tz)
 
@@ -103,7 +106,7 @@ __all__ = [
     "is_timestamp",
     "isstr",
     "iso_to_gregorian",
-    "datetime_from_timestamp",
+    "windows_datetime_from_timestamp",
     "safe_utcfromtimestamp",
     "safe_fromtimestamp",
 ]
--- a/arrow/util.py
+++ b/arrow/util.py
@@ -3,8 +3,8 @@ from __future__ import absolute_import
 
 import datetime
 import math
-import os
 import time
+from os import name as os_name
 
 import dateutil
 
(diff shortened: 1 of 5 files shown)


arrow/util.py
```diff
@@ -23,6 +28,42 @@ def is_timestamp(value):
         return False
 
 
+def datetime_from_timestamp(timestamp, tz=None):
```
systemcatch: Maybe this function could be called `windows_datetime_from_timestamp` to make it clear what it's purpose is.

arrow/util.py
```diff
@@ -23,6 +28,42 @@ def is_timestamp(value):
...
+    """Computes datetime from timestamp. Supports negative timestamps on Windows platform."""
+    sec_frac, sec = math.modf(timestamp)
+    dt = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()) + datetime.timedelta(
+        seconds=sec, microseconds=sec_frac * 1e6
+    )
+    if tz is None:
```
systemcatch: Hmmm I think arrow will generally set timezones to UTC rather than local if no info is provided, @jadchaar any thoughts?

jadchaar: Yeah I think you are correct. This should probably be UTC by default rather than local.

arrow/util.py
kkoziara: Sure, will change that.

arrow/util.py
kkoziara: Ok, cool

arrow/util.py
```diff
@@ -23,6 +28,45 @@ def is_timestamp(value):
...
+                ts
+            ) - datetime.datetime.utcfromtimestamp(ts)
+            dt = dt.replace(tzinfo=dateutil.tz.tzlocal())
+        else:
+            dt = dt.astimezone(tz)
+    return dt
+
+
+def safe_utcfromtimestamp(timestamp, os_name=os.name):
+    """ datetime.utcfromtimestamp alternative which supports negatvie timestamps on Windows platform."""
```
systemcatch: There's a typo in negative here.

arrow/util.py
```diff
@@ -23,6 +28,45 @@ def is_timestamp(value):
...
+    if os_name == "nt" and timestamp < 0:
+        return windows_datetime_from_timestamp(timestamp, dateutil.tz.tzutc())
+    else:
+        return datetime.datetime.utcfromtimestamp(timestamp)
+
+
+def safe_fromtimestamp(timestamp, tz=None, os_name=os.name):
+    """ datetime.fromtimestamp alternative which supports negatvie timestamps on Windows platform."""
```
systemcatch: Same here.

arrow/util.py
```diff
@@ -23,6 +28,45 @@ def is_timestamp(value):
         return False
 
 
+def windows_datetime_from_timestamp(timestamp, tz=None):
+    """Computes datetime from timestamp. Supports negative timestamps on Windows platform."""
+    sec_frac, sec = math.modf(timestamp)
```
jadchaar: My only concern is with the use of `modf`. It seems like it does some pecuiliar rounding:

```shell
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> import math
>>> sec_frac, sec = math.modf(1002130120312.12388128318)
>>> sec_frac
0.1239013671875
>>> sec
1002130120312.0
```

arrow/util.py
```diff
@@ -23,6 +28,45 @@ def is_timestamp(value):
...
+    dt = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()) + datetime.timedelta(
+        seconds=sec, microseconds=sec_frac * 1e6
```
jadchaar: Also it may be better to write `1e6` as `1000000` just for clarity :).

arrow/util.py
```diff
@@ -23,6 +28,45 @@ def is_timestamp(value):
...
+    else:
+        return datetime.datetime.utcfromtimestamp(timestamp)
+
+
+def safe_fromtimestamp(timestamp, tz=None, os_name=os.name):
+    """ datetime.fromtimestamp alternative which supports negatvie timestamps on Windows platform."""
+    if os_name == "nt" and timestamp < 0:
+        if tz is None:
+            # because datetime.fromtimestamp default is local time
+            tz = dateutil.tz.tzlocal()
```
jadchaar: Do we want to set this? According to [this SO response](https://stackoverflow.com/questions/18812638/get-timezone-used-by-datetime-datetime-fromtimestamp), the tzinfo is None by default in `fromtimestamp`.

arrow/util.py
kkoziara: I don't think that's `modf` problem, but rather a python float64 precision problem. If you just store this number it will behave the same.

arrow/util.py
kkoziara: It is None, but the `fromtimestamp` returns datetime adjusted for the local time, so without it `windows_datetime_from_timestamp` won't return proper datetime. I could remove the tz after the call to `windows_datetime_from_timestamp`, but I don't think it spoils anything or am I wrong?

arrow/util.py
jadchaar: I am reading about `modf` on StackOverflow and it seems to have some precision issues: https://stackoverflow.com/q/3886402/.

The post recommends using the decimal library instead:

```shell
❯ python3
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> from decimal import Decimal
>>> dec = Decimal.from_float(1002130120312.12388128318)
>>> dec
Decimal('1002130120312.1239013671875')
>>> neg_dec = Decimal.from_float(-1002130120312.12388128318)
>>> neg_dec
Decimal('-1002130120312.1239013671875')
>>> float(dec % 1)
0.1239013671875
>>> float(neg_dec % 1)
-0.1239013671875
>>> dec.is_signed()
False
>>> neg_dec.is_signed()
True
```

arrow/util.py
jadchaar: Hmm @systemcatch what do you think?

arrow/util.py
jadchaar: Ah I see. You are right, this seems to simply be due to the way that Python stores floats. It feels odd to me because I thought Python3 stores floats as a 64-bit double under the hood, so it seems odd that the precision is lost so early.

I played with a few examples, but this time I stored the large floating point in a variable rather than directly passing it in, and you are correct:
```shell
❯ python3
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> from decimal import Decimal
>>> import math
>>> test_dec = 1002130120312.12388128318
>>> test_dec
1002130120312.1239
>>> str(test_dec)
'1002130120312.1239'
>>> "{:f}".format(test_dec)
'1002130120312.123901'
>>> "{:.32f}".format(test_dec)
'1002130120312.12390136718750000000000000000000'
>>> test_dec = 312.12388128318
>>> test_dec
312.12388128318
>>> math.modf(test_dec)
(0.123881283180026, 312.0)
>>> Decimal.from_float(test_dec)
Decimal('312.123881283180026002810336649417877197265625')
>>> test_dec = 312.123881
>>> math.modf(test_dec)
(0.12388099999998303, 312.0)
>>> Decimal.from_float(test_dec)
Decimal('312.12388099999998303246684372425079345703125')
```

Since timestamps should generally only have 6 decimal places, this minimal loss of precision should be ok then.

jadchaar:
@kkoziara this looks solid, but do you think you can remove these lines in `parser_tests.py`: https://github.com/crsmithdev/arrow/blob/e67c810ba6252b57d0b4504a68c7fd4cf8625f1f/tests/parser_tests.py#L231-L233. We originally commented out these lines because negative time stamps were unsupported on Windows, but now that they are supported, we can run this test on all platforms.

tests/util_tests.py
```diff
@@ -33,3 +35,71 @@ def test_iso_gregorian(self):
...
+        timestamp = -1572204340.6460679
+        result = util.safe_fromtimestamp(timestamp, tz.gettz("Europe/Paris"))
+        expected = datetime(
+            1920, 3, 7, 5, 34, 19, 353932, tzinfo=tz.gettz("Europe/Paris")
+        )
+        self.assertEqual(result, expected)
+
+    def test_safe_utcfromtimestamp_negative_nt(self):
+        timestamp = -1572204340.6460679
+        result = util.safe_utcfromtimestamp(timestamp, os_name="nt").replace(
```
jadchaar: Instead of passing in `os_name` into this function, let's instead use the Python mock library (already in `requirements.txt`) to patch `os.name` to "nt". This way, we can remove the `os_name` argument since it probably won't be used outside of unit tests.

Or, we can only run this test when `os.name == "nt"`.

arrow/util.py
```diff
@@ -23,6 +28,45 @@ def is_timestamp(value):
...
 
 
+def windows_datetime_from_timestamp(timestamp, tz=None):
+    """Computes datetime from timestamp. Supports negative timestamps on Windows platform."""
+    sec_frac, sec = math.modf(timestamp)
+    dt = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()) + datetime.timedelta(
+        seconds=sec, microseconds=sec_frac * 1e6
+    )
+    if tz is not None:
+        if tz == dateutil.tz.tzlocal():
```
jadchaar: @systemcatch and I just reviewed this PR over a call, and we are a bit confused about the logic here.

We tried the following example:

```shell
>>> import arrow
>>> timestamp = -1572204340.6460679
>>> arrow.util.windows_datetime_from_timestamp(timestamp)
1920-03-07 04:34:19.353932+00:00
```

But when we run `datetime.fromtimestamp` (running on macOS), we get the following:
```shell
>>> datetime.fromtimestamp(-1572204340.6460679)
datetime.datetime(1920, 3, 6, 23, 34, 19, 353932)
```

I am on the EST timezone (5 hours behind UTC), and it seems that the UTC tzinfo being set on line 34 is persisting when no timezone is passed in. Should we be converting to `tzlocal()` all the time to align with the behavior of `datetime.fromtimestamp` (ref: https://docs.python.org/3.7/library/datetime.html#datetime.datetime.fromtimestamp).

kkoziara: In a comment from a few days ago you asked me to change `util.windows_datetime_from_timestamp` to return UTC, as it's more aligned with general arrow behaviour. `util.safe_fromtimestamp` is the one emulating the behaviour  of `datetime.fromtimestamp`

tests/util_tests.py
kkoziara: Ok, will try that

arrow/util.py
systemcatch: Hey @kkoziara that comment I made was probably wrong on reflection, we should revert back to the previous behaviour.

jadchaar: Yeah sorry for this @kkoziara, we misspoke. We should maintain the functionality currently present in Arrow 0.15.4, with the addition of Windows negative timestamp support:

```shell
venv ❯ python3
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> import arrow
>>> arrow.__version__
'0.15.4'
>>> timestamp = -1572204340.6460679
>>> arrow.Arrow.fromtimestamp(timestamp)
<Arrow [1920-03-06T23:34:19.353932-05:00]>
>>> arrow.Arrow.utcfromtimestamp(timestamp)
<Arrow [1920-03-07T04:34:19.353932+00:00]>
>>> datetime.fromtimestamp(timestamp)
datetime.datetime(1920, 3, 6, 23, 34, 19, 353932)
>>> datetime.utcfromtimestamp(timestamp)
datetime.datetime(1920, 3, 7, 4, 34, 19, 353932)
```

Also, when parsing timestamps with the "X" token, Arrow seems to explicitly return a UTC timestamp.

arrow/util.py
```diff
@@ -23,6 +28,44 @@ def is_timestamp(value):
...
+def windows_datetime_from_timestamp(timestamp, tz=None):
+    """Computes datetime from timestamp. Supports negative timestamps on Windows platform."""
+    sec_frac, sec = math.modf(timestamp)
+    dt = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()) + datetime.timedelta(
+        seconds=sec, microseconds=sec_frac * 1000000
+    )
+    if tz is None:
+        tz = dateutil.tz.tzlocal()
+
+    if tz == dateutil.tz.tzlocal():
```
jadchaar: Thanks for making the requested changes @kkoziara! This is a great PR and it works well from what I have tried locally!

Before we merge this in, think you can explain the content of this conditional a bit more (lines 40-47)?

kkoziara: Lines 40-47 adjust the `dt` datetime computed using `timedelta` for the local timezone. Normally it's possible to use `datetime.astimezone` method for this, but on Windows for `tz.tzlocal()` and dates before the epoch this does not work. That's why I made this convoluted workaround. Here:

        ts = time.mktime(dt.replace(year=1970).timetuple())
 
I get the non-negative timestamp with the same day of the year as `dt` to have the same daylight saving time as `dt`. Then I use:

        dt += datetime.datetime.fromtimestamp(ts) - datetime.datetime.utcfromtimestamp(
            ts
        )

to get the offset to the UTC for the local timezone and proper daylight saving time, as original `fromtimestamp` will work here (`ts` is non-negative). Then I modify the computed `dt` with the offset. Finally in here: `dt = dt.replace(tzinfo=dateutil.tz.tzlocal())`  we just set the proper `tzinfo`.

I know this is not very straightforward but I didn't get any simpler ideas to achieve this. I am open to change this to something simpler.

jadchaar: I think it is fine--it is a fantastic workaround.

arrow/util.py
```diff
@@ -23,6 +28,44 @@ def is_timestamp(value):
         return False
 
 
+def windows_datetime_from_timestamp(timestamp, tz=None):
+    """Computes datetime from timestamp. Supports negative timestamps on Windows platform."""
+    sec_frac, sec = math.modf(timestamp)
+    dt = datetime.datetime(1970, 1, 1, tzinfo=dateutil.tz.tzutc()) + datetime.timedelta(
+        seconds=sec, microseconds=sec_frac * 1000000
```
jadchaar: This should be good to go, but I had a quick question about this line as well: is it ok that we are passing all of the `sec_frac` into microseconds? Or do we have to explicitly break it up into microseconds and milliseconds when passing the parameters into timedelta?

It should be fine, but I wanted to make sure.

kkoziara: I think in the end there's no difference between passing `x` milliseconds and 1000 * `x` microseconds to the `timedelta`. The `datetime` will be adjusted by the same value.

jadchaar:
Thanks again for the contribution @kkoziara. This is a great PR.

jadchaar:
Hey @kkoziara, I am running the tests on my Mac and noticed that the following test is failing on the current master:

```shell
======================================================================
FAIL: test_windows_datetime_from_timestamp (tests.util_tests.UtilTests)
----------------------------------------------------------------------
  File "/Users/jadchaar/Documents/arrow/.tox/py37/lib/python3.7/site-packages/chai/chai.py", line 63, in wrapper
    func(self, *args, **kwargs)
  File "/Users/jadchaar/Documents/arrow/tests/util_tests.py", line 44, in test_windows_datetime_from_timestamp
    self.assertEqual(result, expected)
AssertionError: datetime.datetime(2019, 10, 27, 14, 25, 40, 646068, tzinfo=tzlocal()) != datetime.datetime(2019, 10, 27, 15, 25, 40, 646068)
```

Seems to be a DST issue. Should this test only be limited to running on Windows?

jadchaar:
@kkoziara any ideas what could be causing this test to fail on master on mac and Linux?

kkoziara:
Hey, sorry for the delay. Have you tested only on mac or also on Linux? I use Linux and haven't seen any problems, but maybe it's somehow related to us having different locale? (it's Warsaw time for me). The method `test_windows_datetime_from_timestamp` is a fallback so in practice it will be only used on Windows. Because it worked on Linux too I just left it to run on all platforms. I think it's ok to limit testing this to Windows only.

PR #705, opened 2019-10-31
Add support for midnight at the end of day

### LINKED ISSUE #703: HH token does not support the number 24, even though the documentation says it should (00 to 24)
According to ISO 8601 `arrow.get('2019-10-29T24:00:00')` should parse to October 30, 2019 (`2019-10-30T00:00:00`), but instead is throwing an exception `ValueError: hour must be in 0..23`

It also fails when doing explicit formatting (same exception)
`arrow.get('2013-12-31T24:00:00', ['YYYY-MM-DD[T]HH:mm:ss'])`

If you look at the docs at https://arrow.readthedocs.io/en/latest/#supported-tokens it says `HH` token is from 00 to 24

if this is accepted as a bug,  it seems a fix for this can be done at  `arrow/parser.py` inside the `_build_datetime` function ( I can do it, including unit tests)

ceorcham:
Fix #703

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -427,6 +427,21 @@ class DateTimeParser(object):
         elif am_pm == "am" and hour == 12:
             hour = 0
 
+        # Support for midnight at the end of day
+        if hour == 24:
+            if parts.get("minute", 0) != 0:
+                raise ParserError("Midnight at the end of day must not contain minutes")
+            if parts.get("second", 0) != 0:
+                raise ParserError("Midnight at the end of day must not contain seconds")
+            if parts.get("microsecond", 0) != 0:
+                raise ParserError(
+                    "Midnight at the end of day must not contain microseconds"
+                )
+            hour = 0
+            day_increment = 1
+        else:
+            day_increment = 0
+
         # account for rounding up to 1000000
         microsecond = parts.get("microsecond", 0)
         if microsecond == 1000000:
@@ -435,7 +450,7 @@ class DateTimeParser(object):
         else:
             second_increment = 0
 
-        increment = timedelta(seconds=second_increment)
+        increment = timedelta(days=day_increment, seconds=second_increment)
 
         return (
             datetime(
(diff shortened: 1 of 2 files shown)


arrow/parser.py
```diff
@@ -423,6 +423,21 @@ def _build_datetime(parts):
         elif am_pm == "am" and hour == 12:
             hour = 0
 
+        # Support for midnight at the end of day
+        if hour == 24:
+            if parts.get("minute"):
```
jadchaar: We should change this and the following conditionals to `if parts.get("minute") is not None:` for readability purposes.

tests/parser_tests.py
```diff
@@ -1176,6 +1176,44 @@ def test_iso8601_basic_format(self):
         with self.assertRaises(ParserError):
             self.parser.parse_iso("20180517T1055213Z")
 
+    def test_midnight_end_day(self):
+        self.assertEqual(
+            self.parser.parse_iso("2019-10-30T24:00:00"),
```
jadchaar: `parse_iso` is only called when `arrow.get` is invoked without a format string (e.g. `arrow.get("2016-05-16")`). We should test this functionality on `self.parser.parse` as well to ensure that functionality like `arrow.get("2016-05-16T24:00:00", "YYYY-MM-DDTHH:mm:ss")` works as expected.

arrow/parser.py
ceorcham: @jadchaar I just made the change, but I had to check against `0` because for this condition is the same if the minute/second/microsecond does not exist or is equal to `0`

tests/parser_tests.py
ceorcham: @jadchaar At the end, both methods `parse_iso` and `parse` use `_build_datetime`, so no extra code changes were needed to support `parse`.
I added the test to ensure it will not break in the future.

ceorcham:
@jadchaar I force-pushed my branch to maintain the same commit message (I also rebased with master branch in the process)

tests/parser_tests.py
jadchaar: Ah yes you are correct! Missed that thanks to the myopic code collapsing in GitHub hah.

arrow/parser.py
jadchaar: Good catch!

PR #714, opened 2019-11-04
Remove extra_requires and use install_requires

ethier:


jadchaar:
Hey @ethier, it seems that your fork is out of date because these changes have already been made on Master (ref: https://github.com/crsmithdev/arrow/blob/master/setup.py#L26).

PR #719, opened 2019-11-21
Allow factory.get() to construct from a timestamp and timezone

systemcatch:
First attempt at this as discussed, not sure about this behaviour though. Any thoughts @jadchaar ?

```shell
[GCC 5.4.0 20160609] on linux
>>> import arrow
>>> arrow.get(1367900664)
<Arrow [2013-05-07T04:24:24+00:00]>
>>> arrow.get(1367900664).replace(tzinfo="US/Pacific")
<Arrow [2013-05-07T04:24:24-07:00]>
>>> from dateutil import tz
>>> time_zone = tz.gettz('US/Pacific')
>>> arrow.get(1367900664, tzinfo=time_zone)
<Arrow [2013-05-06T21:24:24-07:00]>
```

### DIFF
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -168,6 +168,10 @@ class ArrowFactory(object):
             if arg is None:
                 return self.type.utcnow()
 
+            # try (int, float) -> from timestamp with tz
+            elif not isstr(arg) and is_timestamp(arg) and tz is not None:
+                return self.type.fromtimestamp(arg, tzinfo=tz)
+
             # try (int, float) -> utc, from timestamp.
             elif not isstr(arg) and is_timestamp(arg):
                 return self.type.utcfromtimestamp(arg)
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -169,12 +169,11 @@ class ArrowFactory(object):
                 return self.type.utcnow()
 
             # try (int, float) -> from timestamp with tz
-            elif not isstr(arg) and is_timestamp(arg) and tz is not None:
-                return self.type.fromtimestamp(arg, tzinfo=tz)
-
-            # try (int, float) -> utc, from timestamp.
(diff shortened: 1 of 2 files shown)


arrow/factory.py
```diff
@@ -168,6 +168,10 @@ def get(self, *args, **kwargs):
             if arg is None:
                 return self.type.utcnow()
 
+            # try (int, float) -> from timestamp with tz
+            elif not isstr(arg) and is_timestamp(arg) and tz is not None:
+                return self.type.fromtimestamp(arg, tzinfo=tz)
+
             # try (int, float) -> utc, from timestamp.
             elif not isstr(arg) and is_timestamp(arg):
                 return self.type.utcfromtimestamp(arg)
```
jadchaar: We can probably combine these into one, something like this:

```python
elif not isstr(arg) and is_timestamp(arg):
     if tz is None:
           tz = tzutc()
     return self.type.fromtimestamp(arg, tz)
```

systemcatch:
@jadchaar but in my example above the time is different in addition to the timezone, which doesn't seem right.

systemcatch:
Hey @jadchaar I implemented your changes as discussed.

tests/factory_tests.py
```diff
@@ -73,6 +73,16 @@ def test_one_arg_timestamp(self):
         with self.assertRaises((OverflowError, ValueError)):
             self.factory.get(timestamp)
 
+    def test_one_arg_timestamp_with_tzinfo(self):
+
+        int_timestamp = int(time.time())
```
jadchaar: May be better to keep this as a float timestamp: `timestamp = time.time()`. Makes sure to test all aspects of the timestamp.

systemcatch: All done!

PR #720, opened 2019-11-25
Improve parsing of local strings with punctuation

### LINKED ISSUE #701: Locale strings not expanded in `get` format
This code:
```
arrow.get("Montag, 9. September 2019, 16:15-20:00", "dddd, D. MMMM YYYY", locale="de")   
```

works perfectly in `arrow` 1.13.1, but in 1.15.2 I ge an `Exception`t:
```
 Failed to match 'dddd, D. MMMM YYYY' when parsing 'Montag, 9. September 2019, 16:15-20:00'
```
where the pattern is not expanded.

Am I doing something wrong?

andrewchouman:
This should fix #701. I decided to allow the punctuation: `` , . ; : ? ! " ' ` [ ] { } ( ) < > ``.

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -299,8 +299,7 @@ class DateTimeParser(object):
 
         starting_word_boundary = (
             r"(?<!\S\S)"  # Don't have two consecutive non-whitespace characters. This ensures that we allow cases like .11.25.2019 but not 1.11.25.2019 (for pattern MM.DD.YYYY)
-            r"(?<![^\,\.\;\:\?\!\"\'\`\[\]\{\}("
-            r")<>\s])"  # This is the list of punctuation that is ok before the pattern (i.e. "It can't not be these characters before the pattern")
+            r"(?<![^\,\.\;\:\?\!\"\'\`\[\]\{\}\(\)<>\s])"  # This is the list of punctuation that is ok before the pattern (i.e. "It can't not be these characters before the pattern")
             r"(\b|^)"  # The \b is to block cases like 1201912 but allow 201912 for pattern YYYYMM. The ^ was necessary to allow a negative number through i.e. before epoch numbers
         )
         ending_word_boundary = (
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -297,12 +297,18 @@ class DateTimeParser(object):
         # "1998-09-12," is permitted. For the full list of valid punctuation,
         # see the documentation.
 
-        starting_punctuation_bound = (
-            r"(?<!\S\S)(?<![^,.;:?!\"'`\[\]{}(" r")<>\s])(\b|^)"
+        starting_word_boundary = (
+            r"(?<!\S\S)"  # Don't have two consecutive non-whitespace characters. This ensures that we allow cases like .11.25.2019 but not 1.11.25.2019 (for pattern MM.DD.YYYY)
+            r"(?<![^\,\.\;\:\?\!\"\'\`\[\]\{\}("
+            r")<>\s])"  # This is the list of punctuation that is ok before the pattern (i.e. "It can't not be these characters before the pattern")
+            r"(\b|^)"  # The \b is to block cases like 1201912 but allow 201912 for pattern YYYYMM. The ^ was necessary to allow a negative number through i.e. before epoch numbers
         )
-        ending_punctuation_bound = r"(?=[,.;:?!\"'`\[\]{}()<>]?(?!\S))"
+        ending_word_boundary = (
+            r"(?=[\,\.\;\:\?\!\"\'\`\[\]\{\}\(\)\<\>]?"  # Positive lookahead stating that these punctuation marks can appear after the pattern at most 1 time
+            r"(?!\S))"  # Don't allow any non-whitespace character after the punctuation
+        )
         bounded_fmt_pattern = r"{}{}{}".format(
-            starting_punctuation_bound, final_fmt_pattern, ending_punctuation_bound
+            starting_word_boundary, final_fmt_pattern, ending_word_boundary
         )
 
         return tokens, re.compile(bounded_fmt_pattern, flags=re.IGNORECASE)
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -297,7 +297,7 @@ class DateTimeParser(object):
         # "1998-09-12," is permitted. For the full list of valid punctuation,
         # see the documentation.
 
-        starting_punctuation_bound = r"(?<!\S\S)(?<!\s[^,.;:?!\"'`\[\]{}(" \
+        starting_punctuation_bound = r"(?<!\S\S)(?<![^,.;:?!\"'`\[\]{}(" \
                                      r")<>\s])(\b|^)"
         ending_punctuation_bound = r"(?=[,.;:?!\"'`\[\]{}()<>]?(?!\S))"
         bounded_fmt_pattern = r"{}{}{}".format(starting_punctuation_bound,
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -218,6 +218,7 @@ class DateTimeParser(object):
         fmt_tokens, fmt_pattern_re = self._generate_pattern_re(fmt)
 
         match = fmt_pattern_re.search(datetime_string)
+
         if match is None:
             raise ParserMatchError(
                 "Failed to match '{}' when parsing '{}'".format(fmt, datetime_string)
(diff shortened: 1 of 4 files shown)


systemcatch:
Hey @andrewchouman looks like the tests are failing due to `black` having some problems with `parser.py` and `parser_tests.py`.

arrow/parser.py
```diff
@@ -292,11 +293,16 @@ def _generate_pattern_re(self, fmt):
...
-        starting_word_boundary = r"(?<![\S])"
-        ending_word_boundary = r"(?![\S])"
+        # Certain punctuation before or after the target pattern such as
+        # "1998-09-12," is permitted. For the full list of valid punctuation,
+        # see the documentation.
+
+        starting_punctuation_bound = (
+            r"(?<!\S\S)(?<![^,.;:?!\"'`\[\]{}(" r")<>\s])(\b|^)"
+        )
+        ending_punctuation_bound = r"(?=[,.;:?!\"'`\[\]{}()<>]?(?!\S))"
```
jadchaar: It would be great if you could provide a [regex101](regex101.com/) example to walk us through the updated regex.

Also, you should escape the following characters if you are matching them literally in a regex: `. ^ $ * + - ? ( ) [ ] { } \ |`. We escape all symbols just to be safe (add a back slash like `\{`).

arrow/parser.py
```diff
@@ -292,11 +293,16 @@ def _generate_pattern_re(self, fmt):
...
         bounded_fmt_pattern = r"{}{}{}".format(
-            starting_word_boundary, final_fmt_pattern, ending_word_boundary
+            starting_punctuation_bound, final_fmt_pattern, ending_punctuation_bound
```
jadchaar: I would still technically consider this to be a custom word boundary, but with the addition of some punctuation to address edge cases.

tests/parser_tests.py
```diff
@@ -527,6 +530,25 @@ def test_parse_with_extra_words_at_start_and_end_valid(self):
...
         )
 
+        self.assertEqual(
+            self.parser.parse(
+                "Meet me at my house on the my birthday (2019-24-11)", "YYYY-DD-MM"
+            ),
+            datetime(2019, 11, 24),
+        )
+
+        self.assertEqual(
```
jadchaar: I would add a comment to this stating that it fixes the case of a **partial** match that is surrounded by punctuation.

tests/parser_tests.py
```diff
@@ -527,6 +530,25 @@ def test_parse_with_extra_words_at_start_and_end_valid(self):
             datetime(2016, 5, 16, 4, 5, 6, 789120),
         )
 
+        self.assertEqual(
```
jadchaar: Let's place these in a new test function and comment it to indicate that it is a regression test for issue #701.

tests/parser_tests.py
```diff
@@ -527,6 +530,25 @@ def test_parse_with_extra_words_at_start_and_end_valid(self):
...
+            self.parser.parse(
+                "Meet me at my house on the my birthday (2019-24-11)", "YYYY-DD-MM"
+            ),
+            datetime(2019, 11, 24),
+        )
+
+        self.assertEqual(
+            self.parser.parse(
+                "Monday, 9. September 2019, 16:15-20:00", "dddd, D. MMMM YYYY"
```
jadchaar: While this is the exact English translation of the buggy case described in https://github.com/crsmithdev/arrow/issues/701, it is always good to include the original, unaltered problematic test case in the regression tests to make sure that we nailed down the original issue. You can simply add it below this one.

arrow/parser.py
andrewchouman: @jadchaar All great feedback. I'll wait a day or two to implement the changes to see if @systemcatch has anything to add. I'll also provide a regex101 example in the next couple days, but I can explain the regex right now:

> (?<!\S\S)

Don't have two consecutive whitespace characters. This ensures that we allow cases like `11.25.2019.` but not `1.11.25.2019` (for pattern MM.DD.YYYY)


> (?<![^,.;:?!\"'`\[\]{}()<>\s])

This is the list of punctuation that is ok before the pattern. (ie "It can't not be these characters before the pattern")


> (\b|^)

The `\b` is to block cases like `1201912` but allow `201912` for pattern YYYYMM. The `^` was necessary to allow a negative number through i.e. before epoch numbers (a strange edge case).


> (?=[,.;:?!\"'`\[\]{}()<>]?

Positive lookahead stating that these punctuation can appear after the pattern zero or one times. The `?` quantifier can't be used before the pattern due to the way regex compiles or something.


> (?!\S)

Don't allow any non-whitespace character after the punctuation.

jadchaar: > > (?!\S)
> 
> Don't allow any non-whitespace character after the punctuation.

We'd probably want cases like this to pass: `arrow.get("Monday, 9. September 2019,testing", "dddd, D. MMMM YYYY")` since the punctuation acts as a separator. The original purpose of the word boundary was to simply prevent cases like `arrow.get("Monday, 9. September 2019testing", "dddd, D. MMMM YYYY")`, but since we are allowing punctuation to follow, I am not sure if that part of the regex is necessary. Open to your thoughts though.

systemcatch: Hey @andrewchouman I really like your explanation for the new regex, maybe we could incorporate that into the code by using https://docs.python.org/3/library/re.html#re.VERBOSE?

docs/index.rst
```diff
@@ -427,6 +428,17 @@ You can also escape regular expressions by enclosing them within square brackets
     >>> arrow.get("Mon Sep 08   16:41:45   2014", fmt)
     <Arrow [2014-09-08T16:41:45+00:00]>
 
+Punctuation
+~~~~~~~~~~~
+
+You're date formats can be precede and/or be preceded by one character of
+punctuation from the list: :code:``, . ; : ? ! " ' \` [ ] { } ( ) < >``
+
+.. code-block:: python
```
systemcatch: One or two more examples would be helpful here.

docs/index.rst
```diff
@@ -427,6 +428,17 @@ You can also escape regular expressions by enclosing them within square brackets
     >>> arrow.get("Mon Sep 08   16:41:45   2014", fmt)
     <Arrow [2014-09-08T16:41:45+00:00]>
 
+Punctuation
+~~~~~~~~~~~
+
+You're date formats can be precede and/or be preceded by one character of
```
systemcatch: Should be `your` instead. Another idea;

```md
Date formats may be fenced on either side by one punctuation character from the following list.....
```

tests/parser_tests.py
andrewchouman: I put it in factory_tests.py because specifying locale can't be done within parser.py

andrewchouman:
Requested Regex101. Let me know if this isn't enough
[Regular Date](https://regex101.com/r/ucM2y6/1)
[Before Epoch](https://regex101.com/r/xyV4CZ/1)
[YYYYMM](https://regex101.com/r/NtnJRj/1)

arrow/parser.py
jadchaar: Hey @andrewchouman, disregard my previous comment. Chris and I think your assumption of not allowing non-whitespace after the punctuation is correct :).

andrewchouman:
Do the changes look good? @jadchaar @systemcatch

tests/parser_tests.py
```diff
@@ -527,6 +530,38 @@ def test_parse_with_extra_words_at_start_and_end_valid(self):
...
+            self.parser.parse("A date is 11.11.2011.", "DD.MM.YYYY"),
+            datetime(2011, 11, 11),
+        )
+
+        with self.assertRaises(ParserMatchError):
+            self.parser.parse("11.11.2011.1 is not a valid date.", "DD.MM.YYYY")
+
+        with self.assertRaises(ParserMatchError):
+            self.parser.parse(
+                "This date has too many punctuation marks following it (11.11.2011).",
```
jadchaar: Looks good overall, thanks for implementing the feedback.

My only concern is that we are making the parsing constraints too strict here by forcing a single piece of punctuation. Should we maybe allow the case of any number of punctuation marks?

Thoughts @systemcatch?

andrewchouman: I don't think any number would work in my implementation but I can increase it to any finite number you want (say 2 or 3 punctuation marks). I think one is fine; more than two is probably overkill.

jadchaar: I tend to agree, but the only thing that concerns me is that this worked pre 0.15.0 (I chose 0.13.0 for example):
```shell
venv ❯ python3
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 16:52:21)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> import arrow
>>> arrow.__version__
'0.13.0'
>>> arrow.get("This date has too many punctuation marks following it 11.11.2011", "DD.MM.YYYY")
<Arrow [2011-11-11T00:00:00+00:00]>
>>> arrow.get("This date has too many punctuation marks following it (11.11.2011)", "DD.MM.YYYY")
<Arrow [2011-11-11T00:00:00+00:00]>
>>> arrow.get("This date has too many punctuation marks following it (11.11.2011).", "DD.MM.YYYY")
<Arrow [2011-11-11T00:00:00+00:00]>
```

This is definitely an improvement, but for full pre-0.15.0 behavior while still containing improvements, we probably need to add support for any number of punctuation marks. Curious, why would finite numbers work but not infinite (e.g. with the `+` quantifier in regex)?

andrewchouman: I'm not sure if you saw my edit, but I can implement it easily following the date, but preceding doesn't work because [you can't use quantifiers with lookbacks](https://stackoverflow.com/questions/9030305/regular-expression-lookbehind-doesnt-work-with-quantifiers-or).

jadchaar: Oh that is interesting, apologies I missed that. Any idea how we could possibly do this without look behinds?

We can merge this in for sure since it is definitely a good fix, but I am wondering if we could potentially do a follow-up to this using another regex technique that allows for more flexibility.

andrewchouman: I just learned regex to fix this issue, so I'm still a beginner. From my limited research while creating the fix (I was trying to do what you're suggesting at first too), I couldn't find a solution. It's probably still worth making a new issue to see if someone with more regex knowledge can solve it.

A half-solution is just allowing a finite but absurdly high number of punctuation marks.

systemcatch: @jadchaar  I think we merge this in and I'll open a new issue to discuss punctuation handling.

jadchaar: Agreed. Merging!

Issue #716, opened 2019-11-08
Humanize: combine granularities


fabads:
Hello,

is it possible to get a combined results of different granularities, eg. 7 years and 10 months ?

BR.
Fabrice

systemcatch:
Hi @fabads at the moment `humanize()` does not support having multiple levels of granularity.

jadchaar:
That being said though, it would be a great feature to add! Feel free to contribute @fabads, this would be a great addition to the library.

PR #722, opened 2019-11-26
Support for multiple granularities in humanize

hwillard98:
This implements a feature in #716 to use multiple granularities as an input to humanize. 

For example, 
```
>>> later140 = self.now.shift(seconds=8400)
>>> later140.humanize(granularity="minute")
"in 140 minutes" 
>>> later140.humanize(granularity=["hour", "minute"])
"in 2 hours and 20 minutes"
```

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -13,7 +13,6 @@ from datetime import datetime, timedelta
 from datetime import tzinfo as dt_tzinfo
 from math import trunc
 
-import six
 from dateutil import tz as dateutil_tz
 from dateutil.relativedelta import relativedelta
 
@@ -58,11 +57,11 @@ class Arrow(object):
     _ATTRS_PLURAL = ["{}s".format(a) for a in _ATTRS]
     _MONTHS_PER_QUARTER = 3
     _SECS_PER_MINUTE = float(60)
-    _SECS_PER_HOUR = float(3600)  # 60 * 60
-    _SECS_PER_DAY = float(86400)  # 60 * 60 * 24
-    _SECS_PER_WEEK = float(604800)  # 60 * 60 * 24 * 7
-    _SECS_PER_MONTH = float(2635200)  # 60 * 60 * 24 * 30.5
-    _SECS_PER_YEAR = float(31557600)  # 60 * 60 * 24 * 365.25
+    _SECS_PER_HOUR = float(60 * 60)
+    _SECS_PER_DAY = float(60 * 60 * 24)
+    _SECS_PER_WEEK = float(60 * 60 * 24 * 7)
+    _SECS_PER_MONTH = float(60 * 60 * 24 * 30.5)
+    _SECS_PER_YEAR = float(60 * 60 * 24 * 365.25)
 
     def __init__(
         self, year, month, day, hour=0, minute=0, second=0, microsecond=0, tzinfo=None
@@ -948,7 +947,7 @@ class Arrow(object):
                     years = sign * int(max(delta / 31536000, 2))
                     return locale.describe("years", years, only_distance=only_distance)
 
-            elif isinstance(granularity, six.string_types):
+            elif util.isstr(granularity):
                 if granularity == "second":
                     delta = sign * delta
                     if abs(delta) < 2:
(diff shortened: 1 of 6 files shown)


hwillard98:
@jadchaar do you have any comments or feedback for these commits?

jadchaar:
Hey @hwillard98, apologies for the delayed response. I was quite busy this holiday weekend. At first glance, the work looks solid, but I'd like to take a closer look when I get some more free time. I talked to @systemcatch and he said he will review these PRs today as well.

arrow/arrow.py
```diff
@@ -56,6 +57,12 @@ class Arrow(object):
     _ATTRS = ["year", "month", "day", "hour", "minute", "second", "microsecond"]
     _ATTRS_PLURAL = ["{}s".format(a) for a in _ATTRS]
     _MONTHS_PER_QUARTER = 3
+    _SECS_PER_MINUTE = float(60)
+    _SECS_PER_HOUR = float(3600)  # 60 * 60
```
jadchaar: Any reason why you expanded the multiplied values and wrote them in a comment? May be more clear to simply keep the multiplication in place inside of the cast.

arrow/arrow.py
```diff
@@ -937,23 +948,23 @@ def humanize(
                     years = sign * int(max(delta / 31536000, 2))
                     return locale.describe("years", years, only_distance=only_distance)
 
-            else:
+            elif isinstance(granularity, six.string_types):
```
systemcatch: It would be nice to avoid introducing an extra dependency in `six`, maybe there's another way we could do this?

jadchaar: We can use the `isstr` function in `utils.py` or we can import it from `future.utils`.

arrow/arrow.py
hwillard98: I'll expand them in the next commit for clarity

arrow/arrow.py
hwillard98: updated to use utils.isstr, good catch

jadchaar:
@hwillard98 thanks for the quick updates. Mind adding a section to the main documentation (you should add it to this file: https://github.com/crsmithdev/arrow/blob/master/docs/index.rst)?

It should go in the Humanize section: https://arrow.readthedocs.io/en/latest/#humanize

docs/index.rst
```diff
@@ -208,6 +208,17 @@ Or another Arrow, or datetime:
...
+
+.. code-block:: python
+
+    >>> present = arrow.utcnow()
+    >>> future = present.shift(minutes=66)
+    >>> future.humanize(present, granularity="minute")
+    'in 66 minutes'
+    >>> future.humanize(present, granularity=["hour", "minute"])
+    'in an hour and 6 minutes'
+
```
hwillard98: @jadchaar I've added documentation here, happy to elaborate on the examples if we want more detail

hwillard98:
Any other changes requested? I added the documentation in the most recent commit amend

arrow/locales.py
```diff
@@ -78,6 +79,27 @@ def describe(self, timeframe, delta=0, only_distance=False):
...
 
+    def describe_multi(self, timeframes, only_distance=False):
+        """ Describes a delta within multiple timeframes in plain language.
+
+        :param timeframes: a list of string, quantity pairs each representing a timeframe and delta.
+        :param only_distance: return only distance eg: "2 hours 11 seconds" without "in" or "ago" keywords
+        """
+
+        humanized = ""
+        for index in range(len(timeframes)):
```
systemcatch: My only comment here is that it might be beneficial to use [enumerate](https://docs.python.org/3.5/library/functions.html#enumerate) when using the index of a list in a `for` loop. However in this instance the gain is probably minimal.

hwillard98: Good point, I've fixed in most recent amend because I agree it makes the code cleaner

docs/index.rst
```diff
@@ -208,6 +208,17 @@ Or another Arrow, or datetime:
...
+Indicate a specific time granularity (or multiple):
+
+.. code-block:: python
+
+    >>> present = arrow.utcnow()
+    >>> future = present.shift(minutes=66)
+    >>> future.humanize(present, granularity="minute")
+    'in 66 minutes'
+    >>> future.humanize(present, granularity=["hour", "minute"])
+    'in an hour and 6 minutes'
```
jadchaar: Let's add another example of this in action on an `arrow` object in the past and one with the `only_distance` attribute set to true.

arrow/arrow.py
```diff
@@ -962,6 +972,46 @@ def humanize(
                 if trunc(abs(delta)) != 1:
                     granularity += "s"
                 return locale.describe(granularity, delta, only_distance=only_distance)
+
+            else:
+                timeframes = []
+                if "year" in granularity:
+                    years = sign * delta / self._SECS_PER_YEAR
+                    delta = delta % self._SECS_PER_YEAR
```
jadchaar: Changing this to `delta %= self._SECS_PER_YEAR` may be a bit cleaner.

tests/arrow_tests.py
```diff
@@ -1479,9 +1485,63 @@ def test_granularity(self):
...
+            self.now.humanize(later108onlydistance, granularity=["year"]), "3 years ago"
+        )
+        self.assertEqual(
+            self.now.humanize(later108onlydistance, granularity=["month", "week"]),
+            "37 months and 4 weeks ago",
+        )
+        self.assertEqual(
+            self.now.humanize(later108onlydistance, granularity=["year", "second"]),
+            "3 years and seconds ago",
+        )
```
jadchaar: You forgot to set `only_distance=False` in the `humanize` function. With `only_distance` set to true, the first one should be `3 years` since `only_distance` removes the `ago` and `in` (ref: https://arrow.readthedocs.io/en/latest/#arrow.arrow.Arrow.humanize)

arrow/arrow.py
```diff
@@ -962,6 +972,46 @@ def humanize(
...
+                    delta = delta % self._SECS_PER_MINUTE
+                    timeframes.append(["minute", minutes])
+                if "second" in granularity:
+                    seconds = sign * delta
+                    timeframes.append(["second", seconds])
+                if len(timeframes) < len(granularity):
+                    raise AttributeError(
+                        "Invalid level of granularity. Please select between 'second', 'minute', 'hour', 'day', 'week', 'month' or 'year'"
+                    )
+                for index in range(len(timeframes)):
```
jadchaar: Let's change this to `for index, (timeframe, delta) in enumerate(timeframes):` like you did below.

arrow/arrow.py
```diff
@@ -962,6 +972,46 @@ def humanize(
                 if trunc(abs(delta)) != 1:
                     granularity += "s"
                 return locale.describe(granularity, delta, only_distance=only_distance)
+
+            else:
+                timeframes = []
+                if "year" in granularity:
```
jadchaar: Just a minor style comment, but putting spaces between the singular if statements may be cleaner since they are not chained conditionals (e.g. `if...elif...elif...else`).

tests/arrow_tests.py
hwillard98: seems like I forgot to pass only_distance=True into the parameters for humanize in this section so the variable looks mis-named and the output looks wrong. In newest commit I passed in this parameter and updated the asserted output to not have any "ago" or "in" words. The humanize functionality is behaving correctly though

docs/index.rst
hwillard98: Added these examples and some additional documentation in the newest commit

PR #733, opened 2019-12-10
Converted factory.py and parser.py to Cython files and updated setup.py

mashfy13:
## Pull Request Checklist

- [y] 🧪 Added **tests** for changed code.
- [n] 🛠️ All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [n] 📚 Updated **documentation** for changed code.
- [y] ⏩ Code is **up-to-date** with the `master` branch.

## Description of Changes

Cython implementation added to the arrow.get() function. The main change I made was updating some of the variables in the factory.get() function to be defined of type cdef int where it made sense. On top of that I made some changes in parser.py which is now parser.pyx because it seemed like this file had a lot to do with the functions called in the get() function. The changes have been tested using a time function call within the code to see how long it took without the Cython changes and then with the Cython changes. The changes are not perfect and could definitely use some touching up on, but I wasn't entirely sure if I was on the right track in changing the types and data structures in the parser.pyx file. I also added 2 or 3 lines in the setup.py file to add the Cython dependency and added one line at the end of the setup to 'cythonize' the factory and parser files.

### DIFF
--- a/arrow/factory.pyx
+++ b/arrow/factory.pyx
@@ -1,0 +1,314 @@
+# -*- coding: utf-8 -*-
+"""
+Implements the :class:`ArrowFactory <arrow.factory.ArrowFactory>` class,
+providing factory methods for common :class:`Arrow <arrow.arrow.Arrow>`
+construction scenarios.
+
+"""
+
+from __future__ import absolute_import
+
+import calendar
+from datetime import date, datetime
+from datetime import tzinfo as dt_tzinfo
+from time import struct_time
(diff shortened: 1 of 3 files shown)


jadchaar:
Hey, we decided that if we are to go through with Cython support, the entire library should be converted at once. Also, I think there alternatives to simply converting the files to ones with `.pyx` extensions. If you would like to explore this alternative, we'd be happy to take another look.

PR #740, opened 2019-12-17
Add type annotations

isac322:
## Pull Request Checklist

## Description of Changes

Added type annotations to all non test codes.

Used python stub file format to annotate for compatibility with Python 2.

`mypy -p arrow -2 --strict` and `mypy -p arrow --strict` both returns `Success: no issues found in 10 source files` (mypy: 0.750)

Did not test with `pyre` and `pytype` because they both can not handle python 3.8 code.


jadchaar:
Hi @isac322, thanks for putting in the time and effort to work on this. Adding type annotations to Arrow has been on my mind for a while.

That said, I have a couple of comments: I worry that adding the interface (`.pyi`) files may become unmaintainable as future contributors contribute to Arrow since type annotations are pretty new to Python and not many people may be familiar with them. I think it might be better to take an inline approach as shown [here](https://github.com/python/mypy#what-is-mypy) so that users are more enticed to update them when they change code. The only problem with this is that Python 3.5+ is required unless we take the commenting approach specified [here](https://mypy.readthedocs.io/en/latest/python2.html).

That begs the question: should we wait until Arrow drops Python 2 support before adding the Python 3.5+ style type annotations, especially considering that Python 2 is approaching EOL in a few weeks (https://github.com/crsmithdev/arrow/issues/739). Or should we go ahead with the commenting approach to maintain backward compatibility with Python 2.

I do agree though that adding type annotations is a great step for Arrow and will help reduce future bugs.

Once again, thanks for the solid work, I just wanted to start a discussion about the changes.

jadchaar:
Hi @isac322, just wanted to check in on this PR. @systemcatch and I discussed these changes and we decided that it is best to integrate type annotations once Arrow has officially dropped support for Python 2.7 since this will allow us to include more maintainable, inline type annotations.

Since Python 2.7 reached EOL today, we plan on dropping Python 2 support from Arrow in a future v0.16.0 release. We would love to integrate inline type annotations for that release and we were wondering if you would be able to help with that. 

Let us know! Thanks!

isac322:
Hi @jadchaar, first of all, I'm sorry for the delay and thank you for the discussion. Because my company use Python 2 and 3 both yet, personally, it would be good if type annotation stored in separate files 😅.
But since Python 2 is officially dropped, I agree with your decision and I think I can help with the integration.
Please mention me on this PR after dropping Python 2, so I can annotate type on function and local variable too.
Thanks!

systemcatch:
Let's close this for now and once python 2 support is dropped we can move towards inline type checking.

Issue #760, opened 2020-03-03
Replace class setup functions with Pytest fixtures


jadchaar:
Since we recently refactored the tests to use Pytest, it would be nice to convert the existing unittest-style setup/teardown class functions into Pytest-style fixtures. This would lead to better maintainability and compatibility going forward.

An example of the `setup_class(cls)` code that should be replaced: https://github.com/crsmithdev/arrow/blob/master/tests/test_factory.py#L17

krisfremen:
I'm gonna tackle this issue this weekend.

jadchaar:
Great to have you back @krisfremen!

krisfremen:
@jadchaar good to be back, pull request ready for review :)

PR #765, opened 2020-03-29
Convert test setup methods to pytest fixtures

krisfremen:
## Pull Request Checklist

- [x] 📚 Updated **documentation** for changed code.

## Description of Changes
Moving all classes to have pytest fixtures

The lang_locale_fixture is a bit of a hack, but this was the easiest way I thought of instead of changing every single class. Open to suggestions to change it!

Cheers!

Closes: #760

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,7 +2,6 @@
 from datetime import datetime
 
 import pytest
-from dateutil import tz
 
 from arrow import arrow, factory, formatter, locales, parser
 
@@ -28,30 +27,6 @@ def time2013_02_03_fixture(request):
 def time2013_02_15_fixture(request):
     request.cls.datetime = datetime(2013, 2, 15, 3, 41, 22, 8923)
     request.cls.arrow = arrow.Arrow.fromdatetime(request.cls.datetime)
-
-
-@pytest.fixture(scope="class")
-def timedst_fixture(request):
-    request.cls.before_1 = arrow.Arrow(
-        2016, 11, 6, 3, 59, tzinfo=tz.gettz("America/New_York")
-    )
-    request.cls.before_2 = arrow.Arrow(2016, 11, 6, tzinfo=tz.gettz("America/New_York"))
-    request.cls.after_1 = arrow.Arrow(
-        2016, 11, 6, 4, tzinfo=tz.gettz("America/New_York")
-    )
-    request.cls.after_2 = arrow.Arrow(
-        2016, 11, 6, 23, 59, tzinfo=tz.gettz("America/New_York")
-    )
-    request.cls.before_3 = arrow.Arrow(
-        2018, 11, 4, 3, 59, tzinfo=tz.gettz("America/New_York")
-    )
-    request.cls.before_4 = arrow.Arrow(2018, 11, 4, tzinfo=tz.gettz("America/New_York"))
-    request.cls.after_3 = arrow.Arrow(
-        2018, 11, 4, 4, tzinfo=tz.gettz("America/New_York")
-    )
-    request.cls.after_4 = arrow.Arrow(
-        2018, 11, 4, 23, 59, tzinfo=tz.gettz("America/New_York")
-    )
 
 
 @pytest.fixture(scope="class")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -41,11 +41,10 @@ def locales_fixture(request):
 
 @pytest.fixture(scope="class")
 def lang_locale_fixture(request):
+    # As locale test classes are prefixed with Test, we are dynamically getting the locale by the test class name.
+    # TestEnglishLocale -> EnglishLocale
     name = request.cls.__name__[4:]
-    if name == "Locale":
-        request.cls.locale = locales.get_locale_by_class_name("EnglishLocale")
-    else:
-        request.cls.locale = locales.get_locale_by_class_name(name)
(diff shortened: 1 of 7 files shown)


jadchaar:
Looks solid on my first pass, but I was wondering: would it be possible to refactor the fixtures into a single `conftest.py` to reduce potential duplication across files and increase maintainability for the future?

krisfremen:
sure thing!

tests/test_arrow.py
```diff
@@ -504,6 +489,7 @@ def test_strftime(self):
         assert result == self.arrow._datetime.strftime("%Y")
 
 
+@pytest.mark.usefixtures("timedst_fixture")
```
jadchaar: Since this fixture is not used anywhere else and represents a special case, let's move the variables that are currently exposed in the fixture into local variables in the function.

tests/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.fixture(scope="class")
+def locales_fixture(request):
+    request.cls.locales = locales._locales
+
+
+@pytest.fixture(scope="class")
+def lang_locale_fixture(request):
+    name = request.cls.__name__[4:]
```
jadchaar: Hah this is definitely a hack. Let's put a comment to prevent confusion and explain the thought process.

tests/test_locales.py
```diff
@@ -61,11 +58,8 @@ def test_locales(self):
         assert len(locales._locales) > 0
 
 
+@pytest.mark.usefixtures("lang_locale_fixture")
 class TestLocale:
```
jadchaar: Can this potentially be combined into `TestEnglishLocale` to remove the need for having the conditional in the `lang_locale_fixture`?

tests/conftest.py
```diff
@@ -0,0 +1,93 @@
...
+from datetime import datetime
+
+import pytest
+from dateutil import tz
+
+from arrow import arrow, factory, formatter, locales, parser
+
+
+@pytest.fixture(scope="class")
+def utcnow_fixture(request):
```
jadchaar: Nitpicky comment, but I think we should remove the `_fixture` prefix since the decorator above and the `@pytest.mark.usefixtures` calls make it obvious.

tests/test_parser.py
```diff
@@ -17,11 +17,8 @@
...
 
 
+@pytest.mark.usefixtures("parser_fixture")
 class TestDateTimeParser:
-    @classmethod
-    def setup_class(cls):
-        cls.parser = parser.DateTimeParser()
-
     def test_parse_multiformat(self, mocker):
         mocker.patch(
```
jadchaar: Just a quick question: the reason you chose to use `@pytest.mark.usefixtures("parser_fixture")` over passing in the fixture as a function argument was to cut down the changes that needed to be made right?

tests/test_arrow.py
krisfremen: got it.

tests/test_locales.py
krisfremen: yes, indeed. nice one.

tests/test_parser.py
krisfremen: That and it looks cleaner IMHO.

tests/conftest.py
krisfremen: done, also did some renaming to have consistency.

tests/conftest.py
krisfremen: added a comment + example.

tests/test_parser.py
jadchaar: I am wondering if we could add `mocker` to the pytest mark: `@pytest.mark.usefixtures("parser_fixture", "mocker")`. Is it possible to not set the `cls` inside of the fixtures and remove the `self` calls inside of the functions?

tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    request.cls.locale = locales.get_locale_by_class_name(name)
+
+
+@pytest.fixture(scope="class")
+def dt_parser(request):
+    request.cls.parser = parser.DateTimeParser()
+
+
+@pytest.fixture(scope="class")
+def dt_regex(request):
```
jadchaar: Might be better to rename this to `dt_parser_regex`.

tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    request.cls.format_regex = parser.DateTimeParser._FORMAT_RE
+
+
+@pytest.fixture(scope="class")
+def dt_tzinfo(request):
```
jadchaar: Might be better to rename this to `tzinfo_parser`.

systemcatch:
Hey I will take a look at this when I get some time, but I'm not familiar with fixtures so probably won't have much input on this.

tests/test_parser.py
krisfremen: > I am wondering if we could add mocker to the pytest mark: @pytest.mark.usefixtures("parser_fixture", "mocker")

not that I'm aware, pytest-mock only works as a function argument.

> Is it possible to avoid setting the clsdirectly inside of the fixtures (it could allow us remove the self calls inside of the functions)?

if we go the function argument way, yes. this will get rid of the self calls.

jadchaar: This is a good starting point and I agree it may be a bit cleaner than the function approach.

PR #766, opened 2020-03-31
Allow formatting with 'x' token

systemcatch:

There's a failing test which we need to discuss, otherwise this is working ok.

```shell
[GCC 5.4.0 20160609] on linux
>>> import arrow
>>> adt=arrow.utcnow()
>>> adt.format("x")
'1585669870688329'
>>> adt.format("X")
'1585669870'
```

### DIFF
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -16,7 +16,7 @@ class DateTimeFormatter(object):
     # emulated in Python's re library, see https://stackoverflow.com/a/13577411/2701578
 
     _FORMAT_RE = re.compile(
-        r"(\[(?:(?=(?P<literal>[^]]))(?P=literal))*\]|YYY?Y?|MM?M?M?|Do|DD?D?D?|d?dd?d?|HH?|hh?|mm?|ss?|SS?S?S?S?S?|ZZ?Z?|a|A|X|W)"
+        r"(\[(?:(?=(?P<literal>[^]]))(?P=literal))*\]|YYY?Y?|MM?M?M?|Do|DD?D?D?|d?dd?d?|HH?|hh?|mm?|ss?|SS?S?S?S?S?|ZZ?Z?|a|A|X|x|W)"
     )
 
     def __init__(self, locale="en_us"):
@@ -99,6 +99,10 @@ class DateTimeFormatter(object):
 
         if token == "X":
             return str(calendar.timegm(dt.utctimetuple()))
+
+        if token == "x":
+            ts = calendar.timegm(dt.utctimetuple()) + (dt.microsecond / 1000000)
+            return str(int(ts * 1000000))
 
         if token == "ZZZ":
             return dt.tzname()
--- a/tests/test_formatter.py
+++ b/tests/test_formatter.py
@@ -107,7 +107,24 @@ class TestDateTimeFormatterFormatToken:
 
         timestamp = time.time()
         dt = datetime.utcfromtimestamp(timestamp)
+
         assert self.formatter._format_token(dt, "X") == str(int(timestamp))
+
+        # time.time() may return a float with greater than 6 digits of precision
+        # time_parts = str(timestamp).split(".")
+        # if len(time_parts[1]) > 6:
+        #     shortened = time_parts[1][:6]
+        #     truncated_timestamp = time_parts[0] + "." + shortened
+        #     timestamp = float(truncated_timestamp)
+        #
+        # assert self.formatter._format_token(dt, "x") == str(int(timestamp*1000000))
+
+        # TODO failing test for discussion with Jad
+        bad_timestamp = 1585668666.7198436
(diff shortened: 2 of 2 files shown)


systemcatch:
All tests passing is weird as that test I've added specifically fails on my local machine. :thinking:

systemcatch:
This is the test failure I get.
```
====================================== FAILURES ======================================
__________________ TestDateTimeFormatterFormatToken.test_timestamp ___________________

self = <tests.test_formatter.TestDateTimeFormatterFormatToken object at 0x7f3af38b2f90>

    def test_timestamp(self):
    
        timestamp = time.time()
    
        assert self.formatter._format_token(dt, "X") == str(int(timestamp))
    
        # time.time() may return a float with greater than 6 digits of precision
        # time_parts = str(timestamp).split(".")
        # if len(time_parts[1]) > 6:
        #     shortened = time_parts[1][:6]
        #     truncated_timestamp = time_parts[0] + "." + shortened
        #     timestamp = float(truncated_timestamp)
        #
        # assert self.formatter._format_token(dt, "x") == str(int(timestamp*1000000))
    
        # TODO failing test for discussion with Jad
        bad_timestamp = 1585668666.7198436
        dt = datetime.utcfromtimestamp(bad_timestamp)
>       assert self.formatter._format_token(dt, "x") == str(
            int(bad_timestamp * 1000000)
        )
E       AssertionError: assert '1585668666719844' == '1585668666719843'
E         - 1585668666719844
E         ?                ^
E         + 1585668666719843
E         ?                ^

tests/test_formatter.py:125: AssertionError
```

jadchaar:
That is peculiar that CI is passing, but it is failing on your machine. This must be a discrepancy regarding how floating point numbers or integers are represented on different OSes or hardware configurations.

This seems to be occurring because the cast to an int is truncating the `7198436/1e6` into `719843` rather than the rounded version of `719844`. You can probably solve this by rounding the division to 6 digits like this: `ts = calendar.timegm(dt.utctimetuple()) + round(dt.microsecond / 1000000, 6)`.

My only worry about this is that it technically goes against the IEEE-754 rounding we use in the parser.

tests/test_formatter.py
```diff
@@ -107,8 +107,13 @@ def test_timestamp(self):
 
         timestamp = time.time()
         dt = datetime.utcfromtimestamp(timestamp)
+
```
jadchaar, line 110: Was this extra newline intentional?

tests/test_formatter.py
```diff
@@ -107,8 +107,13 @@ def test_timestamp(self):
...
         assert self.formatter._format_token(dt, "X") == str(int(timestamp))
 
+        # time.time() may return a float with greater than 6 digits of precision
+        rounded_ts = str(round(timestamp * 1000000))
+        assert self.formatter._format_token(dt, "x") == rounded_ts.format("{f}")
```
jadchaar, line 115: Any reason you chose to do `self.formatter._format_token(dt, "x") == rounded_ts.format("{f}")` over just `self.formatter._format_token(dt, "x") == rounded_ts`?

arrow/formatter.py
```diff
@@ -100,6 +100,10 @@ def _format_token(self, dt, token):
         if token == "X":
             return str(calendar.timegm(dt.utctimetuple()))
 
+        if token == "x":
+            ts = calendar.timegm(dt.utctimetuple()) + (dt.microsecond / 1000000)
+            return str(int(ts * 1000000))
```
jadchaar, line 105: Do we have to worry about rounding >6 digits of precision here as well?

tests/test_formatter.py
systemcatch: Nope, removing

tests/test_formatter.py
systemcatch: Without that format it defaults to an exponential style which causes an AssertionError.

```
E       AssertionError: assert '1585764864000000' == '1.5857648648e+15'
E         - 1585764864000000
E         + 1.5857648648e+15
```

arrow/formatter.py
systemcatch: Not here since `datetime` microseconds is capped at 6 digits.

tests/test_formatter.py
jadchaar: Oh interesting. Did not realize that :D.

arrow/formatter.py
jadchaar: Let's put a test in our suite to test this edge case.

jadchaar: Ah in that case I don't think a test is necessary then.

systemcatch:
Do you think we need any docs changes for this?

jadchaar:
> Do you think we need any docs changes for this?

Good question. I think it may have been covered when we added the parsing support: https://arrow.readthedocs.io/en/latest/#supported-tokens.

Issue #762, opened 2020-03-04
Add common format strings


jadchaar:
## Feature Request

Pendulum has a nice feature that allows you to make a function call on a Pendulum object to automatically get a pre-formatted, common string (e.g. a cookie time string). See this page for examples: https://pendulum.eustace.io/docs/#common-formats.

rbandrews3:
Hi, I am interested in working on this issue. I forked the repository, cloned that repo, and ran "pip install -U arrow". It seems to successfully install, but when I try to run some of the test files, I get various errors with import packages. I am wondering if this is normal, or if we have done something wrong in setup. 

_Examples of some issues:_
"Traceback (most recent call last):
  File "tests/test_parser.py", line 17, in <module>
    from .utils import make_full_tz_list
ImportError: attempted relative import with no known parent package"

"Traceback (most recent call last):
  File "test_factory.py", line 12, in <module>
    from .utils import assert_datetime_equality
ModuleNotFoundError: No module named '__main__.utils'; '__main__' is not a package"

My friend, @ikoningstein, is working on this with me. Thank you in advance!

jadchaar:
Hey @rbandrews3, by doing `pip install -U Arrow`, you are installing the packaged version from PyPi. Since you are developing the package, you need to simply execute the pytest test cases like this:

```shell
make build37 # replace with whatever python version you have on your system
pytest -v tests # this assumes you are in the root arrow directory
```

Or, you can use Tox (recommended) since this handles creating the virtual environment and installing the dependencies for you:
```shell
pip install -U tox
tox
```
The `tox` step runs linting and runs the tests on available python versions that are installed on your system.

ikoningstein:
Thanks so much! We got the library running smoothly now. @rbandrews3  and I read through the documentation on the Pendulum common formatting, and want to confirm that this feature requires the implementation of the following functions. 

- dt.to_atom_string()
- dt.to_cookie_string()
- dt.to_iso8601_string()
- dt.to_rfc822_string()
- dt.to_rfc850_string()
- dt.to_rfc1036_string()
- dt.to_rfc1123_string()
- dt.to_rfc2822_string()
- dt.to_rfc3339_string()
- dt.to_rss_string()
- dt.to_w3c_string()

Thanks!

jadchaar:
You do not have to implement all of them if you do not want to, so you can make a call on which of the functions you would like to implement based on the workload. These functions will most likely live in the Arrow formatter`format.py`.

[...]

PR #770, opened 2020-04-06
Added cookie, rfc822, rfc850, and rfc1036 to built-in formats

rbandrews3:
## Pull Request Checklist

- [X] 🧪 Added **tests** for changed code.
- [X] 🛠️ All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [X] ⏩ Code is **up-to-date** with the `master` branch.

## Description of Changes

We added 4 functions to arrow/arrow/formatter.py to format a DateTime object to support cookie, rfc822, rfc850, rfc1036 formats. We added a test for each function to arrow/tests/test_formatter.py. 

This is for issue #762. We plan to continue implementing more formats based on the Pendulum documentation provided in the issue description after receiving feedback on our first 4.

@ikoningstein and I look forward to hearing your feedback.

### DIFF
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -120,3 +120,79 @@ class DateTimeFormatter(object):
         if token == "W":
             year, week, day = dt.isocalendar()
             return "{}-W{:02d}-{}".format(year, week, day)
+
+
+    # COOKIE FORMAT: 'Thursday, 25-Dec-1975 14:15:16 EST'
+    def _format_cookie(self, dt):
+        # Date information
+        weekday = self._format_token(dt, "dddd")
+        day = self._format_token(dt, "DD")
+        month = self._format_token(dt, "MMM")
+        year = self._format_token(dt, "YYYY")
+        
+        # Time information 
+        hour = self._format_token(dt, "HH")
+        minute = self._format_token(dt, "mm")
+        second = self._format_token(dt, "ss")
+        timezone = self._format_token(dt, "ZZZ")
+        
+        day_format = weekday + ", " + day + "-" + month + "-" + year + " " 
+        time_format = hour + ":" + minute + ":" + second + " " + timezone
+        return day_format + time_format
+
+
+    # rfc822 FORMAT: 'Thu, 25 Dec 75 14:15:16 -0500'
+    def _format_rfc822(self, dt):
+        # Date information
+        weekday = self._format_token(dt, "ddd")
+        day = self._format_token(dt, "DD")
+        month = self._format_token(dt, "MMM")
+        year = self._format_token(dt, "YY")
+        
+        # Time info
+        hour = self._format_token(dt, "HH")
+        minute = self._format_token(dt, "mm")
+        second = self._format_token(dt, "ss")
+        timezone = self._format_token(dt, "Z")
+        
+        day_format = weekday + ", " + day + " " + month + " " + year + " " 
+        time_format = hour + ":" + minute + ":" + second + " " + timezone
+        return day_format + time_format
+
+
+    # rfc850 FORMAT: 'Thursday, 25-Dec-75 14:15:16 EST'
+    def _format_rfc850(self, dt):
+        # Date information
+        weekday = self._format_token(dt, "dddd")
+        day = self._format_token(dt, "DD")
+        month = self._format_token(dt, "MMM")
+        year = self._format_token(dt, "YY")
+        
+        # Time information 
+        hour = self._format_token(dt, "HH")
+        minute = self._format_token(dt, "mm")
+        second = self._format_token(dt, "ss")
+        timezone = self._format_token(dt, "ZZZ")
+        
+        day_format = weekday + ", " + day + "-" + month + "-" + year + " " 
+        time_format = hour + ":" + minute + ":" + second + " " + timezone
+        return day_format + time_format
+
+
+    #  rfc1036 FORMAT: 'Thu, 25 Dec 75 14:15:16 -0500'
+    def _format_rfc1036(self, dt):
+        # Date information
+        weekday = self._format_token(dt, "ddd")
+        day = self._format_token(dt, "DD")
+        month = self._format_token(dt, "MMM")
+        year = self._format_token(dt, "YY")
+        
+        # Time info
+        hour = self._format_token(dt, "HH")
+        minute = self._format_token(dt, "mm")
+        second = self._format_token(dt, "ss")
+        timezone = self._format_token(dt, "Z")
+        
+        day_format = weekday + ", " + day + " " + month + " " + year + " " 
+        time_format = hour + ":" + minute + ":" + second + " " + timezone
+        return day_format + time_format
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -123,16 +123,16 @@ class DateTimeFormatter(object):
 
     # COOKIE FORMAT: 'Thursday, 25-Dec-1975 14:15:16 EST'
     def format_cookie(cls, dt):
-        return cls.format(dt, 'dddd, DD-MMM-YYYY HH:mm:ss ZZZ')
+        return cls.format(dt, "dddd, DD-MMM-YYYY HH:mm:ss ZZZ")
 
     # rfc822 FORMAT: 'Thu, 25 Dec 75 14:15:16 -0500'
     def format_rfc822(cls, dt):
-        return cls.format(dt, 'ddd, DD MMM YY HH:mm:ss Z')
+        return cls.format(dt, "ddd, DD MMM YY HH:mm:ss Z")
 
     # rfc850 FORMAT: 'Thursday, 25-Dec-75 14:15:16 EST'
     def format_rfc850(cls, dt):
-        return cls.format(dt, 'dddd, DD-MMM-YY HH:mm:ss ZZZ')
+        return cls.format(dt, "dddd, DD-MMM-YY HH:mm:ss ZZZ")
 
     #  rfc1036 FORMAT: 'Thu, 25 Dec 75 14:15:16 -0500'
     def format_rfc1036(cls, dt):
-        return cls.format(dt, 'ddd, DD MMM YY HH:mm:ss Z')
+        return cls.format(dt, "ddd, DD MMM YY HH:mm:ss Z")
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -121,7 +121,6 @@ class DateTimeFormatter(object):
             year, week, day = dt.isocalendar()
             return "{}-W{:02d}-{}".format(year, week, day)
 
-
     # COOKIE FORMAT: 'Thursday, 25-Dec-1975 14:15:16 EST'
     def _format_cookie(self, dt):
         # Date information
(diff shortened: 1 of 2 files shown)


krisfremen:
Hi @rbandrews3, code looks good so far.

We use black for code formatting as part of our CI, it tends to be strict. It seems like some of your code isn't up to its strict standards.

Easily fixable though, black will do all the work. just execute:
```pre-commit run --all-files```
afterwards you can commit and push the fixes for the linting CI to pass.

Cheers.

File: arrow/formatter.py [hunk 1]
```diff
@@ -120,3 +120,79 @@ def _format_token(self, dt, token):
...
+        
+        # Time information 
+        hour = self._format_token(dt, "HH")
+        minute = self._format_token(dt, "mm")
+        second = self._format_token(dt, "ss")
+        timezone = self._format_token(dt, "ZZZ")
+        
+        day_format = weekday + ", " + day + "-" + month + "-" + year + " " 
+        time_format = hour + ":" + minute + ":" + second + " " + timezone
+        return day_format + time_format
```
jadchaar, line 141: Let's use a format string similar to the one used on line 122.

tests/test_formatter.py
```diff
@@ -203,3 +203,83 @@ def test_escape(self):
 
         # Escaping is atomic: brackets inside brackets are treated literally
         assert self.formatter.format(datetime(1, 1, 1), "[[[ ]]") == "[[ ]"
+    
+
+    # COOKIE FORMAT: 'Thursday, 25-Dec-1975 14:15:16 EST'
```
jadchaar, line 208: Let's put these tests into their own class called `TestDateTimeFormatterPremadeFormats`.

tests/test_formatter.py
```diff
@@ -203,3 +203,83 @@ def test_escape(self):
...
+    def test_cookie(self):
+        for full_name in make_full_tz_list():
+            # Testing functionality of basic values and timezones
+            dt = datetime(1975, 12, 25, 14, 15, 16, tzinfo=pytz.timezone("UTC")).replace(
+                tzinfo=dateutil_tz.gettz(full_name)
+            )
+            abbreviation = dt.tzname()
+            correct_timezone = self.formatter._format_token(dt, "ZZZ")
+            assert self.formatter._format_cookie(dt) == "Thursday, 25-Dec-1975 14:15:16 " + correct_timezone
```
jadchaar, line 217: Let's use format strings here as well.

rbandrews3:
Hi @krisfremen and @jadchaar,

Thank you for the feedback. @ikoningstein and I made the changes you both told us to and have pushed again. Please let us know if there are any other adjustments to be made!

arrow/formatter.py
```diff
@@ -120,3 +120,75 @@ def _format_token(self, dt, token):
         if token == "W":
             year, week, day = dt.isocalendar()
             return "{}-W{:02d}-{}".format(year, week, day)
+
+    # COOKIE FORMAT: 'Thursday, 25-Dec-1975 14:15:16 EST'
+    def _format_cookie(self, dt):
```
jadchaar, line 125: So these functions are meant to be used like this:

```
>>> ar = arrow.get("1975-12-25T14:15:16")
<Arrow [1975-12-25T14:15:16+00:00]>
>>> ar.format_cookie()
'Thursday, 25-Dec-1975 14:15:16 EST'
```

We need to pass in `cls` rather than `self` and we should remove the leading underscores.

tests/test_formatter.py
```diff
@@ -203,3 +203,100 @@ def test_escape(self):
...
         assert self.formatter.format(datetime(1, 1, 1), "[[[ ]]") == "[[ ]"
+
+
+class TestDateTimeFormatterPremadeFormats:
+    def setup_class(cls):
+        cls.formatter = formatter.DateTimeFormatter()
+
+    # cookie format: 'Thursday, 25-Dec-1975 14:15:16 EST'
+    def test_cookie(self):
+        for full_name in make_full_tz_list():
```
systemcatch, line 214: There's no need to test against the full tz list here, since those tests already happen in `test_timezone_formatter`.

tests/test_formatter.py
```diff
@@ -203,3 +203,100 @@ def test_escape(self):
...
+
+            # Testing that single digit values return with a 0 in front
+            dt_single_digit = datetime(
+                1975, 2, 5, 4, 5, 6, tzinfo=pytz.timezone("UTC")
+            ).replace(tzinfo=dateutil_tz.gettz(full_name))
+            correct_timezone = self.formatter._format_token(dt_single_digit, "ZZZ")
+            assert (
+                self.formatter._format_cookie(dt_single_digit)
+                == "Wednesday, 05-Feb-1975 04:05:06 " + correct_timezone
+            )
```
systemcatch, line 233: This is a good test but again we only need one timezone here.

tests/test_formatter.py
```diff
@@ -203,3 +203,100 @@ def test_escape(self):
...
+
+    # rfc833 format: 'Thu, 25 Dec 75 14:15:16 -0500'
+    def test_rfc_822(self):
+        for full_name in make_full_tz_list():
```
systemcatch, line 237: Same comment as above.

tests/test_formatter.py
```diff
@@ -203,3 +203,100 @@ def test_escape(self):
...
+            correct_timezone = self.formatter._format_token(dt_single_digit, "Z")
+            assert (
+                self.formatter._format_rfc822(dt_single_digit)
+                == "Wed, 05 Feb 75 04:05:06 " + correct_timezone
+            )
+
+    # rfc850 format: 'Thursday, 25-Dec-75 14:15:16 EST'
+    def test_rfc_850(self):
+        for full_name in make_full_tz_list():
+            # Testing functionality of basic values and timezones
```
systemcatch, line 261: Same as above re: timezones, we also don't really need this comment.

tests/test_formatter.py
```diff
@@ -203,3 +203,100 @@ def test_escape(self):
...
+            correct_timezone = self.formatter._format_token(dt_single_digit, "ZZZ")
+            assert (
+                self.formatter._format_rfc850(dt_single_digit)
+                == "Wednesday, 05-Feb-75 04:05:06 " + correct_timezone
+            )
+
+    #  rfc1036 format: 'Thu, 25 Dec 75 14:15:16 -0500'
+    def test_rfc_1036(self):
+        for full_name in make_full_tz_list():
+            # Testing functionality of basic values and timezones
```
systemcatch, line 284: See above

tests/test_formatter.py
```diff
@@ -203,3 +203,100 @@ def test_escape(self):
...
+            correct_timezone = self.formatter._format_token(dt, "Z")
+            assert (
+                self.formatter._format_rfc1036(dt)
+                == "Thu, 25 Dec 75 14:15:16 " + correct_timezone
+            )
+
+            # Testing that single digit values return with a 0 in front
+            dt_single_digit = datetime(
+                1975, 2, 5, 4, 5, 6, tzinfo=pytz.timezone("UTC")
+            ).replace(tzinfo=dateutil_tz.gettz(full_name))
```
systemcatch, line 297: Since we aren't testing vs the whole tz database, you can define the timezone straight away without having to replace.

arrow/formatter.py
```diff
@@ -120,3 +120,75 @@ def _format_token(self, dt, token):
...
+        year = self._format_token(dt, "YYYY")
+
+        # Time information
+        hour = self._format_token(dt, "HH")
+        minute = self._format_token(dt, "mm")
+        second = self._format_token(dt, "ss")
+        timezone = self._format_token(dt, "ZZZ")
+
+        return "{}, {}-{}-{} {}:{}:{} {}".format(
+            weekday, day, month, year, hour, minute, second, timezone
```
krisfremen, line 139: @rbandrews3 good work, would be nice to have this in a more succinct form.

Instead of doing every single one of the returns with a string format, arrow also has the format function, to which you can pass an entire formatted string and converts the tokens appropriately.

https://arrow.readthedocs.io/en/latest/#format

```arrow.utcnow().format('YYYY-MM-DD HH:mm:ss ZZ')```

Cheers!

File: tests/test_formatter.py [hunk 12]
```diff
@@ -203,3 +203,100 @@ def test_escape(self):
 
         # Escaping is atomic: brackets inside brackets are treated literally
         assert self.formatter.format(datetime(1, 1, 1), "[[[ ]]") == "[[ ]"
+
+
+class TestDateTimeFormatterPremadeFormats:
```
krisfremen, line 208: We recently switched and merged in changes to the tests, mainly moving them all to pytest and using fixtures, can you rebase your changes and have this class use the formatter fixture instead?

arrow/formatter.py
jadchaar: Yeah it might be good to do something similar to what we do in the `format` function itself: https://github.com/crsmithdev/arrow/blob/master/arrow/formatter.py#L28.

This way we can easily allow for translations to be inserted into the formatted time strings.

tests/test_formatter.py
rbandrews3: Thank you for your feedback! Unfortunately, we are having trouble understanding how to go about rebasing our code because we are working in our own branch on a forked repository, while the fixture updates are in the master branch. Do you have any guidance as to how we should go about merging these changes?

rbandrews3:
Thank you for all the feedback. @ikoningstein and I made updates to our use of the format function and the timezones in our testing. In regard to the timezones, we aren't sure if what we did is completely correct, so please let us know what other updates might need to be made.

tests/test_formatter.py
krisfremen: https://github.com/crsmithdev/arrow/blob/master/tests/test_formatter.py#L12

Put up this decorator on top of the class. and you can remove the setup_class function.

tests/test_formatter.py
```diff
@@ -203,3 +203,87 @@ def test_escape(self):
...
+    def setup_class(cls):
+        cls.formatter = formatter.DateTimeFormatter()
+
+    # cookie format: 'Thursday, 25-Dec-1975 14:15:16 EST'
+    def test_cookie(self):
+        tz = list(make_full_tz_list())[0]
```
systemcatch, line 214:
```suggestion
        tz = "America/New_York"
```
Since we only need one timezone let's avoid generating the whole list. This change can be replicated for each format.

tests/test_formatter.py
```diff
@@ -203,3 +203,87 @@ def test_escape(self):
...
+        dt = datetime(1975, 12, 25, 14, 15, 16, tzinfo=dateutil_tz.gettz(tz))
+
+        correct_timezone = self.formatter._format_token(dt, "ZZZ")
+        assert (
+            self.formatter.format_cookie(dt)
+            == "Thursday, 25-Dec-1975 14:15:16 " + correct_timezone
+        )
```
systemcatch, line 221:
```suggestion
        assert (
            self.formatter.format_cookie(dt)
            == "Thursday, 25-Dec-1975 14:15:16 EST"
        )
```
Not particularly important just a suggestion, feel free to leave these parts as they are.

tests/test_formatter.py
```diff
@@ -203,3 +203,87 @@ def test_escape(self):
...
+
+        # Testing that single digit values return with a 0 in front
+        dt_single_digit = datetime(
+            1975, 2, 5, 4, 5, 6, tzinfo=pytz.timezone("UTC")
+        ).replace(tzinfo=dateutil_tz.gettz(tz))
```
systemcatch, line 226:
```suggestion
            1975, 2, 5, 4, 5, 6, tzinfo=dateutil_tz.gettz(tz)
        )
```
No need to replace, we can start with the right tz straight away.

rbandrews3:
We have implemented all of the changes. If we can do anything else before implementing the rest of the suggested pendulum formats, let us know. Thank you!

systemcatch:
Hey @rbandrews3 just checking in to say we haven't forgotten your PR, work and exams are getting in the way right now.

We're discussing how to make these functions more usable, arrow objects have access to the `.format()` method which is easy to use. At the moment the built in formats have to be used like below which is awkward.

```shell
[GCC 5.4.0 20160609] on linux
>>> import arrow
>>> arw=arrow.utcnow()
>>> arw
<Arrow [2020-04-20T19:03:28.817475+00:00]>
>>> fmtter=arrow.formatter.DateTimeFormatter()
>>> fmtter.format_cookie(arw)
'Monday, 20-Apr-2020 19:03:28 UTC'
```

If you can bear with us we should have a decision soon.

Issue #360, opened 2016-10-08
Add support for reversed humanize


paulschmeida:
I think it would be nice if there was a reversed version of humanize, i.e. a function that parses human readable input like "20 hours ago" and produces an Arrow object than can be used in scripts with ease.

ju2pom:
Exactly what I need !

andrewelkins:
#361 would address this I think

systemcatch:
Despite the linked PR now being closed, this is still something we'd like to implement. I think we could do with a good idea for structure + spec before we PR this again though.

abschmidt6:
I'm interested in working on this feature, I will take a look at it!

jadchaar:
Fantastic. Let us know if you have any questions.

sawhitt:
I am interested in working on this, would that be alright?

jadchaar:
Sure go for it @sawhitt !

systemcatch:
#772 was a good try at implementing this functionality but it failed to account for any locales other than English. Future attempts at this should take into account that many languages don't separate with spaces (Chinese, Japanese) and don't always have a value followed by a unit (Swedish).

svanita00:
Hi! I'm interested in working on this issue. I'll be taking a look at it for the next few weeks.

anish-nya:
I am interested working on this issue.

systemcatch:
Now that #956 has been merged I think we can close this.

PR #772, opened 2020-04-18
Dehumanize function implementation and tests

sawhitt:
## Pull Request Checklist

- [x] 📚 Updated **documentation** for changed code.

## Description of Changes
Added implementation for a dehumanize function that addresses the want in #360 .
This is currently a very simple implementation with chances to be expanded upon.

Closes: #360

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -1084,10 +1084,10 @@ class Arrow(object):
         times = timestring.split(" ")
 
         if times[-1] == "ago":
-            shiftVal = -1
+            sign = -1
             times = times[:-1]
         elif times[0] == "in":
-            shiftVal = 1
+            sign = 1
             times = times[1:]
         else:
             raise ValueError("Invalid prefix or suffix")
@@ -1095,6 +1095,11 @@ class Arrow(object):
         if len(times) % 2 != 0:
             raise ValueError("Invalid time input")
 
+        second = 0
+        minute = 0
+        hour = 0
+        day = 0
+        week = 0
         month = 0
         year = 0
 
@@ -1103,32 +1108,31 @@ class Arrow(object):
             unit = times[i + 1]
 
             if unit in ["second", "seconds"]:
-                current += shiftVal * timedelta(seconds=val)
+                second = sign * val
             elif unit in ["minute", "minutes"]:
-                current += shiftVal * timedelta(minutes=val)
+                minute = sign * val
             elif unit in ["hour", "hours"]:
-                current += shiftVal * timedelta(hours=val)
+                hour = sign * val
             elif unit in ["day", "days"]:
-                current += shiftVal * timedelta(days=val)
+                day = sign * val
             elif unit in ["week", "weeks"]:
-                current += shiftVal * timedelta(weeks=val)
+                week = sign * val
             elif unit in ["month", "months"]:
-                month += shiftVal * val
+                month = sign * val
             elif unit in ["year", "years"]:
-                year += shiftVal * val
+                year = sign * val
             else:
                 raise ValueError("Error parsing time string")
 
-        month = current.month + month
-        year += month // 12
-        year = current.year + year
-        month = month % 12
-
-        dateobj = datetime(
-            year, month, current.day, current.hour, current.minute, current.second
+        return current.shift(
+            seconds=second,
+            minutes=minute,
+            hours=hour,
+            days=day,
+            weeks=week,
+            months=month,
+            years=year,
         )
-
-        return self.fromdatetime(dateobj)
 
     # query functions
 
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -1056,6 +1056,78 @@ class Arrow(object):
                     e, locale_name
                 )
             )
+    
+    def dehumanize(self, timestring):
+        """ Returns an arrow object relative to the humanized difference in time given.
+            Only works with specifically formatted strings for the time being.
+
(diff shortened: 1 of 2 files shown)


jadchaar:
Hey @sawhitt it seems like the linting CI tests are failing. You can fix it by running the following locally:

```shell
pip install -U pre-commit
pre-commit install
pre-commit run --all-files
```

Or
```shell
pip install -U tox
tox -e lint
```

sawhitt:
Just fixed, sorry about that!

arrow/arrow.py
```diff
@@ -1057,6 +1057,79 @@ def humanize(
...
+        :param timestring: A string in the format returned by humanize by default
+            Number must be before unit of time, i.e. (4 hours)
+
+        Usage:
+
+            >>> arrow.utcnow()
+            <Arrow [2020-04-17T00:57:03+00:00]>
+
+            >>> arrow.utcnow().dehumanize("4 hours ago")
+            <Arrow [2020-04-17T20:57:03+00:00]>
```
systemcatch, line 1073: It would be nice if this didn't require an existing object kind of like how [humanize](https://github.com/crsmithdev/arrow/blob/master/arrow/arrow.py#L866) works currently, i.e.

```shell
>>> arrow.dehumanize("in 4 hours")
<Arrow [2020-04-19T23:58:23+00:00]>
```
This provides flexibility while allowing the user to select a different "now" if required.

arrow/arrow.py
```diff
@@ -1057,6 +1057,79 @@ def humanize(
...
+            >>> arrow.utcnow().dehumanize("4 days 7 hours 10 minutes 5 seconds ago")
+            <Arrow [2020-04-12T19:46:58+00:00]>
+        """
+
+        current = self.fromdatetime(self._datetime)
+
+        times = timestring.split(" ")
+
+        if times[-1] == "ago":
+            shiftVal = -1
```
systemcatch, line 1087: Generally variables in python don't use camel case, how about calling this `sign` instead?

arrow/arrow.py
```diff
@@ -1057,6 +1057,79 @@ def humanize(
...
+                raise ValueError("Error parsing time string")
+
+        month = current.month + month
+        year += month // 12
+        year = current.year + year
+        month = month % 12
+
+        dateobj = datetime(
+            year, month, current.day, current.hour, current.minute, current.second
+        )
```
systemcatch, line 1129: I think (not fully tested though) that we could use arrow's [shift](https://github.com/crsmithdev/arrow/blob/master/arrow/arrow.py#L634) method here to avoid having to calculate the change manually.

arrow/arrow.py
sawhitt: Like making dehumanize a class function instead?

sawhitt:
I just pushed a new version which changed the variable name and now uses shift as suggested by @systemcatch . I have not changed the function to be a class function for arrow yet. From what I see humanize requires an object to use, are you suggesting to make dehumanize and class function that makes an object relative to the current time?

systemcatch:
@sawhitt hmmm maybe I need to rethink how we want this method to work. `humanize` does need an object to work.

```shell
[GCC 5.4.0 20160609] on linux
>>> import arrow
>>> dt=arrow.utcnow()
>>> dt
<Arrow [2020-04-20T16:30:11.604258+00:00]>
>>> dt.humanize()
'just now'
>>> early=dt.shift(hours=-3)
>>> early
<Arrow [2020-04-20T13:30:11.604258+00:00]>
>>> arrow.Arrow.humanize(early)
'3 hours ago'
```

Calling `dehumanize` on an object seems redundant given that it's already in a non-human form (@krisfremen any thoughts here?).

arrow/arrow.py
```diff
@@ -1057,6 +1057,83 @@ def humanize(
...
+
+        return current.shift(
+            seconds=second,
+            minutes=minute,
+            hours=hour,
+            days=day,
+            weeks=week,
+            months=month,
+            years=year,
+        )
```
systemcatch, line 1135:
```suggestion
        val = int(times[0])
        unit = times[1]
        
        shifted = current.shift(**{unit: sign*val})

        return self.fromdatetime(shifted)
```

arrow/arrow.py
```diff
@@ -1057,6 +1057,83 @@ def humanize(
...
+            sign = -1
+            times = times[:-1]
+        elif times[0] == "in":
+            sign = 1
+            times = times[1:]
+        else:
+            raise ValueError("Invalid prefix or suffix")
+
+        if len(times) % 2 != 0:
+            raise ValueError("Invalid time input")
```
systemcatch, line 1096: Let's make these error messages more detailed, with a prompt on how the user can fix the problem.

tests/test_arrow.py
```diff
@@ -1297,6 +1297,65 @@ def test_bounds_are_validated(self):
...
+        arw = arrow.Arrow(2014, 4, 18, 12, 30, 0)
+
+        with pytest.raises(ValueError):
+            arw.dehumanize("4 hours")
+
+        with pytest.raises(ValueError):
+            arw.dehumanize("in 4 hours 3")
+
+        with pytest.raises(ValueError):
+            arw.dehumanize("in 3 lightyears")
```
systemcatch, line 1356: :smile:

arrow/arrow.py
systemcatch: Ahh we'll have to make sure the unit is plural before being passed to `shift`.

PR #773, opened 2020-04-19
Skip dateparser tests when the package is not installed

mgorny:
## Pull Request Checklist

## Description of Changes

This PR makes the dateparser tests skipped if dateparser is not installed. This would help us since we don't have it packaged and don't want yet another package to maintain for one test.

### DIFF
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -2,7 +2,6 @@
 import time
 from datetime import date, datetime
 
-import dateparser
 import pytest
 from dateutil import tz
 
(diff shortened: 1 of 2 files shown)


tests/test_factory.py
```diff
@@ -2,14 +2,18 @@
...
 
 from arrow.parser import ParserError
 
 from .utils import assert_datetime_equality
 
+try:
+    import dateparser
+except ImportError:
+    dateparser = None
+
```
jadchaar, line 16: Thanks for bringing this up. This is a great use case for pytest importorskip: https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency. Mind replacing this implementation with importorskip and we can get this merged in?

mgorny: Thanks, I didn't know about that (or expect such a thing to exist ;-)).

jadchaar: Hah me neither! I just had a hunch that something like this existed so I did some googling and came across it!

Issue #465, opened 2017-10-02
`datetime.timestamp` returns `float`, but Arrow's returns `int`


Syeberman:
As mentioned in https://github.com/crsmithdev/arrow/pull/464 (thanks @zmwangx!), Python 3.3 added a `datetime.timestamp()` method returning a `float` with microsecond precision.  Unfortunately, Arrow already has an `Arrow.timestamp` **property**, which is an `int` with second precision.

As `Arrow` is intended to be a drop-in replacement for `datetime`, my opinion is that this forces us to implement a `timestamp` **method** as well.  We can implement this and maintain backwards compatibility if the `timestamp` property returned an `int`-like object that, when called, returned a float.  Accessing it like a property could be accompanied with a `DeprecationWarning` so we could remove this hack eventually.

jadchaar:
Hi @Syeberman, I am looking to fix this issue and was wondering where to start. I was looking at `arrow.py` and noticed that there is a `float_timestamp` method, and was wondering if this is as simple as creating a method called `timestamp()` and simply mimicking the functionality of `float_timestamp` (reducing duplicate code of course). Any input is appreciated 😄 .

zmwangx:
Here's some proof of concept code to implement the original proposal, namely "the timestamp property returned an int-like object that, when called, returned a float", as is:

```py
class Timestamp(int):
    def __new__(cls, arrow_obj: Arrow, *args, **kwargs):
        instance = int.__new__(cls, calendar.timegm(arrow_obj._datetime.utctimetuple()), *args, **kwargs)
        instance._arrow = arrow_obj
        return instance

    def __call__(self) -> float:
        return self._arrow.float_timestamp


class Arrow(...):
    ...
    
    @property
    def timestamp(self):
        return Timestamp(self)
```

The "accessing it like a property could be accompanied with a DeprecationWarning" part is harder since in Python 3 implicit lookup of magic methods bypass `__getattribute__` completely, so adding something like the following to the `Timestamp` class

```py
    def __getattribute__(self, name):
        if name not in ["_arrow", "__call__"]:
            warnings.warn("Arrow.timestamp as a property (int) is deprecated", DeprecationWarning)
        return object.__getattribute__(self, name)
```

won't work when the `Timestamp` is used like a regular int (with implicit `__str__`, `__add__`, etc.). I can't think of a solution without overriding all magic methods of `int`.

To be clear, I'm not advocating this solution. In fact, I think it (at least this particular implementation) is batshit crazy.

[...]

PR #814, opened 2020-06-26
Add new int_timestamp property in preparation for 1.0.0 release

systemcatch:
## Pull Request Checklist

- [ ] 🛠️ All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [ ] 🧹 All linting checks **pass** when run locally (run `tox -e lint` or `make lint` to find out!).
- [ ] ⏩ Code is **up-to-date** with the `master` branch.

## Description of Changes

This is the start of implementing #465 

- New property `int_timestamp` that does exactly the same thing as `.timestamp`.
- `timestamp` will be removed in 1.0.0
- I have not touched `float_timestamp` in this PR.
- `float_timestamp` essentially does the same as `datetime.timestamp()`, we could leave it as a legacy property.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -694,7 +694,8 @@ class Arrow(object):
 
         warnings.warn(
             "For compatibility with the datetime.timestamp() method this property will be replaced with a method in "
-            "the 1.0.0 release, please switch to .int_timestamp for identical behaviour as soon as possible.",
+            "the 1.0.0 release, please switch to the .int_timestamp property for identical behaviour as soon as "
+            "possible.",
             DeprecationWarning,
         )
         return calendar.timegm(self._datetime.utctimetuple())
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -693,14 +693,23 @@ class Arrow(object):
         """
 
         warnings.warn(
-            "For datetime compatibility reasons this property will be removed in the future, instead use "
-            ".unix_timestamp for the same behaviour",
-            PendingDeprecationWarning,
+            "For compatibility with the datetime.timestamp() method this property will removed in the 1.0.0 release, "
+            "please switch to .int_timestamp for identical behaviour as soon as possible.",
+            DeprecationWarning,
         )
         return calendar.timegm(self._datetime.utctimetuple())
 
     @property
-    def unix_timestamp(self):
+    def int_timestamp(self):
+        """Returns a timestamp representation of the :class:`Arrow <arrow.arrow.Arrow>` object, in
+        UTC time.
+
+        Usage::
+
+            >>> arrow.utcnow().timestamp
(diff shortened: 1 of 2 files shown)


arrow/arrow.py
```diff
@@ -692,6 +692,25 @@ def timestamp(self):
...
+        return calendar.timegm(self._datetime.utctimetuple())
+
+    @property
+    def int_timestamp(self):
+        """Returns a timestamp representation of the :class:`Arrow <arrow.arrow.Arrow>` object, in
+        UTC time.
+
+        Usage::
+
+            >>> arrow.utcnow().timestamp
```
jadchaar, line 709: Comment needs to change.

arrow/arrow.py
```diff
@@ -692,6 +692,25 @@ def timestamp(self):
 
         """
 
+        warnings.warn(
+            "For compatibility with the datetime.timestamp() method this property will removed in the 1.0.0 release, "
```
jadchaar, line 696: Should we instead say "this property will be replaced with a function in 1.0.0"?

arrow/arrow.py
jadchaar, line 709: Comment needs to change.

jadchaar (changes requested):
Couple of tweaks.

arrow/arrow.py
```diff
@@ -692,6 +692,25 @@ def timestamp(self):
 
         """
 
+        warnings.warn(
+            "For compatibility with the datetime.timestamp() method this property will be replaced with a method in "
+            "the 1.0.0 release, please switch to .int_timestamp for identical behaviour as soon as possible.",
```
jadchaar, line 697: Should we say "please switch to the int_timestamp property as soon as possible"? This way it is clear that we are replacing a property with another property?

arrow/arrow.py
```diff
@@ -706,7 +725,11 @@ def float_timestamp(self):
 
         """
 
-        return self.timestamp + float(self.microsecond) / 1000000
+        # IDEA get rid of this in 1.0.0 and wrap datetime.timestamp()
+        # Or for compatibility retain this but make it call the timestamp method
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
```
jadchaar, line 731: This code is meant to catch the deprecation warning since it should not emit on this method correct?

jadchaar (approved):
Couple of comments, but looks good!

arrow/arrow.py
systemcatch: Yeah since the `float_timestamp` property calls `.timestamp` internally we need to suppress the warning but ensure the filter is restored afterwards.

jadchaar (approved):
Great, looks good to merge!

PR #828, opened 2020-07-21
Adding a deprecation warning for Python 2.7 and <=3.5.

krisfremen:
## Pull Request Checklist

- [x] 🧹 All linting checks **pass** when run locally (run `tox -e lint` or `make lint` to find out!).

## Description of Changes

Adding a deprecation warning for Python 2.7 and <=3.5.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -19,6 +19,15 @@ from dateutil.relativedelta import relativedelta
 
 from arrow import formatter, locales, parser, util
 
+if sys.version_info[:2] < (3, 6):  # pragma: no cover
+    with warnings.catch_warnings():
+        warnings.simplefilter("default", DeprecationWarning)
+        warnings.warn(
+            "Arrow will drop support for Python 2.7 and 3.5 in the upcoming v1.0.0 release. Please upgrade to "
+            "Python 3.6+ to continue receiving updates for Arrow.",
+            DeprecationWarning,
+        )
+
 
 class Arrow(object):
     """An :class:`Arrow <arrow.arrow.Arrow>` object.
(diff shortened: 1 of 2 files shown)


arrow/arrow.py
```diff
@@ -66,6 +67,14 @@ class Arrow(object):
     def __init__(
         self, year, month, day, hour=0, minute=0, second=0, microsecond=0, tzinfo=None
     ):
+        if sys.version_info[:2] < (3, 6):  # pragma: no cover
+            with warnings.catch_warnings():
+                warnings.simplefilter("default")
```
jadchaar, line 72: Do we need these? Don't we just need to emit the warning and let the default warning behavior handle the rest?

arrow/arrow.py
```diff
@@ -66,6 +67,14 @@ class Arrow(object):
...
+                warnings.warn(
+                    "Arrow will drop support for Python 2.6 and 3.5 in the upcoming v1.0.0 release. Please upgrade to "
```
jadchaar, line 74: This should be 2.7.

tests/test_arrow.py
```diff
@@ -1805,3 +1805,12 @@ def test_get_iteration_params(self):
...
             arrow.Arrow._get_iteration_params(None, None)
+
+
+class TestArrowPy2Deprecation:
+    @pytest.mark.skipif(
+        sys.version_info[:2] > (3, 5), reason="requires python3.6 or higher"
+    )
+    def test_py2deprecation(self):
+        with pytest.deprecated_call():
+            arrow.Arrow(2020, 12, 12)
```
jadchaar, line 1816: Let's test 3.6+ to make sure the warning is NOT emitted. This will ensure the coverage gets to 100%.

arrow/arrow.py
systemcatch: The way python handles these warnings is kinda counter intuitive.

`default` means that the warning will be shown. DeprecationWarning only shows when the module is run as `__main__` on 3.7+, it won't show on 2.7/3.5/3.6 unless we force it.

Kris is using the `catch_warnings` context manager to make sure we reset the warning filter after we modify it. We should probably make it specific to `DeprecationWarning` though.

`warning.simplefilter("default", DeprecationWarning)`

Issue #842, opened 2020-08-23
Range gives inconsistent results for month end


jadchaar:
## Issue Description

```shell
env ❯ python3
Python 3.8.5 (default, Jul 21 2020, 10:48:26)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import arrow
>>> from datetime import datetime
>>> start_date = datetime(2014, 12, 31)
>>> list(arrow.Arrow.range("month", start_date, limit=5))
[<Arrow [2014-12-31T00:00:00+00:00]>, <Arrow [2015-01-31T00:00:00+00:00]>, <Arrow [2015-02-28T00:00:00+00:00]>, <Arrow [2015-03-28T00:00:00+00:00]>, <Arrow [2015-04-28T00:00:00+00:00]>
```

When months in a range end on different dates (e.g. January and February), the range seems to continue with the end date of the previous month.

## System Info

- 🖥 **OS name and version**: macOS 10.15.6 
- 🐍 **Python version**: 3.8.5 
- 🏹 **Arrow version**: 0.15.8

jadchaar:
This issue may occur as well with years since leap years can increase the number of days in a year.

systemcatch:
Fixed in #844

PR #844, opened 2020-08-28
Fix range behaviour at end of month/quarter/year frames (#842)

JonathanMeans:
## Pull Request Checklist

- [x] 🧹 All linting checks **pass** when run locally (run `tox -e lint` or `make lint` to find out!).

## Description of Changes

Thanks for the library. This PR provides a fix for #842

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -385,6 +385,106 @@ class Arrow(object):
                 **{frame_relative: relative_steps}
             )
 
+    def span(self, frame, count=1, bounds="[)"):
+        """ Returns two new :class:`Arrow <arrow.arrow.Arrow>` objects, representing the timespan
+        of the :class:`Arrow <arrow.arrow.Arrow>` object in a given timeframe.
+
+        :param frame: the timeframe.  Can be any ``datetime`` property (day, hour, minute...).
+        :param count: (optional) the number of frames to span.
+        :param bounds: (optional) a ``str`` of either '()', '(]', '[)', or '[]' that specifies
+            whether to include or exclude the start and end values in the span. '(' excludes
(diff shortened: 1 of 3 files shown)


systemcatch:
Hey @JonathanMeans thanks for the PR, we've not forgotten it. All 3 of us are really busy right now.

arrow/arrow.py
```diff
@@ -1534,6 +1542,10 @@ def _get_iteration_params(cls, end, limit):
                 return end, sys.maxsize
             return end, limit
 
+    @classmethod
+    def _is_last_day_of_month(cls, date):
+        return (date + timedelta(days=1)).day == 1
```
systemcatch, line 1547: Since this doesn't use `cls` at all maybe it should be a `@staticmethod`?

arrow/arrow.py
```diff
@@ -1567,6 +1575,10 @@ def _get_iteration_params(cls, end, limit):
                 return end, sys.maxsize
             return end, limit
 
+    @staticmethod
+    def _is_last_day_of_month(date):
+        return (date + timedelta(days=1)).day == 1
```
jadchaar, line 1580: I am thinking it might be better to use the calendar module to handle this like in the following SO answer: https://stackoverflow.com/a/53591311/3820660. Adding a time delta might get dicey in the case of a DST transition. Using shift is also a possibility since that should account for imaginary datetimes (ref @systemcatch).

systemcatch: Using `calendar` should help us avoid the dst and imaginary issues.

jadchaar (approved):
Brilliant PR. Thanks for the work @JonathanMeans. I will let @systemcatch have the final word here, but it looks great to me!

systemcatch (approved):

Issue #861, opened 2020-10-02
Remove Python 2.7/3.5-specific pragma or skipif statements


jadchaar:
There are a few places in our codebase where we ignore code coverage (`pragma: no cover`) due to skipping tests on older Python versions. We need to remove these along with skipif statements that are specific to Python 2.7/3.5.

Issue #739, opened 2019-12-16
Get ready for Python 2 EOL


systemcatch:
Python 2 is reaching EOL in 2020 (https://pythonclock.org/).

Me and @jadchaar have chatted about this before but haven't come to a decision yet. We'd hoped that the DST upgrades would be done before the EOL so Python 2.7 users would at least have access to those improvements. However due to being busy we've not had time to work on that.

An important factor in this decision will be how long travis-ci supports 2.7.

dateutil discussion - https://github.com/dateutil/dateutil/issues/653

@hugovk do you have any thoughts or stats on Python 2.7 usage?

hugovk:
Here's the pip installs for arrow from PyPI for November 2019:

| category | percent | downloads |
|----------|--------:|----------:|
| 3.7      |  36.63% |   722,467 |
| 3.6      |  24.36% |   480,459 |
| 2.7      |  24.06% |   474,529 |
| 3.5      |  12.03% |   237,263 |
| 3.8      |   1.46% |    28,852 |
| null     |   0.88% |    17,421 |
| 3.4      |   0.57% |    11,185 |
| 2.6      |   0.00% |        86 |
| 3.3      |   0.00% |        47 |
| 3.9      |   0.00% |         5 |
| Total    |         | 1,972,314 |

Date range: 2019-11-01 - 2019-11-30

Source: `pip install -U pypistats && pypistats python_minor arrow --last-month`

---

Compare with some other projects:

* https://dev.to/hugovk/python-version-share-over-time-5-7ik

24% is fairly low, and in itself shouldn't be an impediment to dropping.

See also NEP 29, a recommendation for Scientific Python projects to support Python 3.6 and higher:

* https://numpy.org/neps/nep-0029-deprecation_policy.html

If the DST upgrades aren't too far off, no problem in waiting a bit. But I wouldn't wait too long and it might be easier to do them having dropped 2.7.

In general, I'd recommend dropping 2.7 sometime around the EOL date, having made one last release with whatever is unreleased.

systemcatch:
Thank you Hugo for the stats and analysis which is very helpful.

I was a little suprised at the 24% still on 2.7 but it seems that is similar or below the usage for other major projects.

We will probably release 0.15.5 before Christmas with a warning about Python 2 deprecation and drop support in 0.16.0 on a date to be determined in this issue.

jadchaar:
[...]

PR #877, opened 2020-11-08
Deprecate Python 2.7 and Python 3.5

jadchaar:
# Python 2.7 & 3.5 Deprecation Checklist

- [x] Update `setup.py` classifiers and `python_requires` to declare arrow only supports Python 3.6+
- [x] Drop Python < 3.6 from CI
- [x] Remove Python < 3.6 from Makefile
- [x] Run `pyupgrade --py36-plus`
- [x] Convert all formatting to f-strings (should be done by pyupgrade)
- [x] Remove all future imports
- [x] Remove any remaining python 2 import shadowing
- [x] Audit and overhaul requirements file
- [x] Change timestamp formatting to use `datetime.timestamp()`
- [x] Update timestamp calls in formatter.py
- [x] Replace hardcoded `MAX_TIMESTAMP` constant with `datetime.max.timestamp()`
- [x] Convert `isinstance(value, numbers.Integral)` to `isinstance(value, int)`
- [x] Remove `isstr` function
- [x] Remove `__cmp__()` method in [arrow.py](https://github.com/arrow-py/arrow/blob/2cfb2b0f010ba60d3d9894f8ba5187b909fa97e8/arrow/arrow.py#L1217-L1222)
- [x] Remove coverage pragmas and skipped tests for py27
- [x] Revisit `enfold` calls in code @systemcatch 

# Breaking changes for v1.0.0

- [x] Change timestamp from property to method in `arrow.py` (requires deprecation warning beforehand)
- [x] Change the return value of Arrow's timestamp method (currently a property) to a float

# General Changes for v1.0.0

- [x] Remove beta label from `setup.py`

Closes: https://github.com/arrow-py/arrow/issues/465
Closes: https://github.com/arrow-py/arrow/issues/861
Closes: https://github.com/arrow-py/arrow/issues/739


jadchaar:
Should be ready for review @systemcatch @krisfremen. All what remains is updating the enfold calls, which @systemcatch is more familiar with.

.pre-commit-config.yaml
```diff
@@ -2,12 +2,12 @@ default_language_version:
   python: python3
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v3.3.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: fix-encoding-pragma
-        exclude: ^arrow/_version.py
```
systemcatch, line 10: This is meant to be gone?

arrow/arrow.py
```diff
@@ -720,7 +697,7 @@ def int_timestamp(self):
 
         """
 
-        return calendar.timegm(self._datetime.utctimetuple())
+        return int(self.timestamp())
```
systemcatch, line 700: Probably best to stick with old way to guarantee compatibility.

arrow/arrow.py
```diff
@@ -734,11 +711,7 @@ def float_timestamp(self):
 
         """
 
-        # IDEA get rid of this in 1.0.0 and wrap datetime.timestamp()
-        # Or for compatibility retain this but make it call the timestamp method
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", DeprecationWarning)
-            return self.timestamp + float(self.microsecond) / 1000000
+        return self.timestamp()
```
systemcatch, line 714: I agree with retaining this.

arrow/parser.py
```diff
@@ -1,19 +1,12 @@
...
 
 from dateutil import tz
 
 from arrow import locales
 from arrow.util import iso_to_gregorian, next_weekday, normalize_timestamp
 
-try:
-    from functools import lru_cache
-except ImportError:  # pragma: no cover
-    from backports.functools_lru_cache import lru_cache  # pragma: no cover
```
systemcatch, line 15: Good to see this go!

setup.cfg
```diff
@@ -1,2 +0,0 @@
-[bdist_wheel]
-universal = 1
```
systemcatch, line 2: No longer needed?

systemcatch (changes requested):
Hey @jadchaar just a few comments but otherwise it looks very solid.

The `enfold` calls aren't an issue on python 3.6+, I suggest we leave them for a later PR as the users won't see much difference.

.pre-commit-config.yaml
jadchaar: Yeah that was there since we didn't want to put the encoding pragma in `_version.py`, but in a Python 3.6+ codebase, the encoding pragma is not needed at all.

setup.cfg
jadchaar: I think it is only needed if we are distributing a wheel for Python 2 and 3: https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

jadchaar:
Thanks for the feedback. We had a note to revisit enfold in the copy constructor since we would now be able to just copy the datetime object as is rather than having to enfold again. Do we want to ignore that and just leave it for a future PR?

systemcatch:
> Thanks for the feedback. We had a note to revisit enfold in the copy constructor since we would now be able to just copy the datetime object as is rather than having to enfold again. Do we want to ignore that and just leave it for a future PR?

Not sure what you mean by copy constructor?

jadchaar:
> > Thanks for the feedback. We had a note to revisit enfold in the copy constructor since we would now be able to just copy the datetime object as is rather than having to enfold again. Do we want to ignore that and just leave it for a future PR?
> 
> Not sure what you mean by copy constructor?

Oh oops sorry I meant `replace`:
https://github.com/arrow-py/arrow/blob/609b630b9f5a4b814377053965b93d78fd979c64/arrow/arrow.py#L777-L823

We probably don't need enfold here anymore since datetime would replace fold natively. Thoughts?

systemcatch:
You're right, how about i make an issue to update the enfold calls after 1.0.0 is released?

jadchaar:
Ref: https://github.com/arrow-py/arrow/issues/878

Want to approve so we can get this merged @systemcatch?

systemcatch (approved):

jadchaar:
@isac322 You should be good to now implement type checking off of master!

PR #883, opened 2020-11-18
Add type annotations

isac322:
## Pull Request Checklist

## Description of Changes

- Added type annotation to all functions and methods.
- Add runtime dependency [typing_extensions](https://pypi.org/project/typing-extensions/) for backward compatibility with 3.6 and 3.7
- Add mypy git pre-commit hook
- Add mypy configuration file (setup.cfg)
- Refactor some logics for better type annotation
  - I will create comments for these

Please review this with my comments on PR. I have questions.


arrow/api.py
```diff
@@ -4,14 +4,74 @@
...
+
+@overload
+def get(
+    __arg1: str,
+    __arg2: Union[str, List[str]],
+    *,
+    locale: str = "en_us",
+    tzinfo: Optional[TZ_EXPR] = None,
+    normalize_whitespace: bool = False,
+) -> Arrow:
```
isac322, line 70: [Positional Only Argument](https://www.python.org/dev/peps/pep-0570/) is perfect for this. but is new on 3.8. So it would be good if you can use the syntax when we deprecate 3.7

arrow/arrow.py
```diff
@@ -714,28 +829,26 @@ def float_timestamp(self):
...
 
     @property
-    def fold(self):
+    def fold(self) -> int:
         """ Returns the ``fold`` value of the :class:`Arrow <arrow.arrow.Arrow>` object. """
 
-        # in python < 3.6 _datetime will be a _DatetimeWithFold if fold=1 and a datetime with no fold attribute
-        # otherwise, so we need to return zero to cover the latter case
-        return getattr(self._datetime, "fold", 0)
+        return self._datetime.fold
```
isac322, line 835: Remove backward compatibility code.

arrow/arrow.py
```diff
@@ -1058,55 +1177,38 @@ def humanize(
...
+                            timeframes.append((_granularity, value))
+                    return _delta
+
+                delta = gather_timeframes(delta, "year")
+                delta = gather_timeframes(delta, "month")
+                delta = gather_timeframes(delta, "week")
+                delta = gather_timeframes(delta, "day")
+                delta = gather_timeframes(delta, "hour")
+                delta = gather_timeframes(delta, "minute")
+                _ = gather_timeframes(delta, "second")
```
isac322, line 1204: Refactoring

arrow/arrow.py
```diff
@@ -1390,84 +1497,101 @@ def for_json(self):
...
 
-    def __add__(self, other):
+    def __add__(self, other: Union[timedelta, relativedelta]) -> "Arrow":
 
         if isinstance(other, (timedelta, relativedelta)):
             return self.fromdatetime(self._datetime + other, self._datetime.tzinfo)
 
-        return NotImplemented
+        # FIXME: change to raise
+        return NotImplemented  # type: ignore
```
isac322, line 1506: why this method returns `NotImplemented` class? I think we should raise `NotImplemented`

arrow/arrow.py
```diff
@@ -1390,84 +1497,101 @@ def for_json(self):
...
-        elif isinstance(other, datetime):
+        elif isinstance(other, dt_datetime):
             return self._datetime - other
 
         elif isinstance(other, Arrow):
             return self._datetime - other._datetime
 
-        return NotImplemented
+        # FIXME: change to raise
+        return NotImplemented  # type: ignore
```
isac322, line 1533: why this method returns `NotImplemented` class? I think we should raise `NotImplemented`

arrow/arrow.py
```diff
@@ -1390,84 +1497,101 @@ def for_json(self):
...
-    def __rsub__(self, other):
+    def __rsub__(self, other: dt_datetime) -> timedelta:
 
-        if isinstance(other, datetime):
+        if isinstance(other, dt_datetime):
             return other - self._datetime
 
-        return NotImplemented
+        # FIXME: change to raise
+        return NotImplemented  # type: ignore
```
isac322, line 1541: why this method returns `NotImplemented` class? I think we should raise `NotImplemented`. or I should change return type to `Union["Arrow", Type[NotImplemented]]`.

arrow/arrow.py
```diff
@@ -1390,84 +1497,101 @@ def for_json(self):
...
         return not self.__eq__(other)
 
-    def __gt__(self, other):
+    def __gt__(self, other: Union["Arrow", dt_datetime]) -> bool:
 
-        if not isinstance(other, (Arrow, datetime)):
-            return NotImplemented
+        if not isinstance(other, (Arrow, dt_datetime)):
+            # FIXME: change to raise
+            return NotImplemented  # type: ignore
```
isac322, line 1563: why this method returns `NotImplemented` class? I think we should raise `NotImplemented`. or I should change return type to `Union["Arrow", Type[NotImplemented]]`.

arrow/arrow.py
```diff
@@ -1390,84 +1497,101 @@ def for_json(self):
...
         return self._datetime > self._get_datetime(other)
 
-    def __ge__(self, other):
+    def __ge__(self, other: Union["Arrow", dt_datetime]) -> bool:
 
-        if not isinstance(other, (Arrow, datetime)):
-            return NotImplemented
+        if not isinstance(other, (Arrow, dt_datetime)):
+            # FIXME: change to raise
+            return NotImplemented  # type: ignore
```
isac322, line 1571: why this method returns `NotImplemented` class? I think we should raise `NotImplemented`. or I should change return type to `Union["Arrow", Type[NotImplemented]]`.

arrow/arrow.py
```diff
@@ -1390,84 +1497,101 @@ def for_json(self):
...
         return self._datetime >= self._get_datetime(other)
 
-    def __lt__(self, other):
+    def __lt__(self, other: Union["Arrow", dt_datetime]) -> bool:
 
-        if not isinstance(other, (Arrow, datetime)):
-            return NotImplemented
+        if not isinstance(other, (Arrow, dt_datetime)):
+            # FIXME: change to raise
+            return NotImplemented  # type: ignore
```
isac322, line 1579: why this method returns `NotImplemented` class? I think we should raise `NotImplemented`. or I should change return type to `Union["Arrow", Type[NotImplemented]]`.

arrow/arrow.py
```diff
@@ -1390,84 +1497,101 @@ def for_json(self):
...
         return self._datetime < self._get_datetime(other)
 
-    def __le__(self, other):
+    def __le__(self, other: Union["Arrow", dt_datetime]) -> bool:
 
-        if not isinstance(other, (Arrow, datetime)):
-            return NotImplemented
+        if not isinstance(other, (Arrow, dt_datetime)):
+            # FIXME: change to raise
+            return NotImplemented  # type: ignore
```
isac322, line 1587: why this method returns `NotImplemented` class? I think we should raise `NotImplemented`. or I should change return type to `Union["Arrow", Type[NotImplemented]]`.

arrow/formatter.py
```diff
@@ -116,7 +123,10 @@ def _format_token(self, dt, token):
         if token in ["ZZ", "Z"]:
             separator = ":" if token == "ZZ" else ""
             tz = dateutil_tz.tzutc() if dt.tzinfo is None else dt.tzinfo
-            total_minutes = int(util.total_seconds(tz.utcoffset(dt)) / 60)
+            # FIXME: utcoffset() is nullable
+            total_minutes = int(
+                util.total_seconds(cast(timedelta, tz.utcoffset(dt))) / 60
+            )
```
isac322, line 129: I think we should refactor this code.
`tz.utcoffset(dt)` is nullable

arrow/arrow.py
jadchaar: I agree, I think this should be a `raise`. This is a good example of some ancient code bloat that has carried on with Arrow over the years.

jadchaar: Actually, it turns out that this should actually be `return NotImplemented`: https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror

arrow/arrow.py
jadchaar: Seems like returning is correct: https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror

arrow/arrow.py
jadchaar: Seems like returning is correct: https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror

arrow/formatter.py
jadchaar: Go for it! Have anything in mind?

arrow/arrow.py
isac322: Hmmm Ok. I'll figure out how to annotate on this situation.

arrow/locales.py
```diff
@@ -92,7 +146,11 @@ def describe(self, timeframe, delta=0, only_distance=False):
 
         return humanized
 
-    def describe_multi(self, timeframes, only_distance=False):
+    def describe_multi(
+        self,
+        timeframes: Collection[Tuple[TimeFrames, Union[int, float]]],
+        only_distance: bool = False,
+    ) -> str:
```
isac322, line 153: I think we should restrict `timeframes` parameter to `Collection[Tuple[TimeFrames, Union[int, float]]]`.
Currently this method can accept arguments something like `[['seconds', 1, 1], (1, 1, 1)]`. Obviously no one use this method like that, but it would be good we can catch error on status analysis.

arrow/locales.py
```diff
@@ -850,13 +920,19 @@ class FinnishLocale(Locale):
 
     day_abbreviations = ["", "ma", "ti", "ke", "to", "pe", "la", "su"]
 
-    def _format_timeframe(self, timeframe, delta):
+    # TODO: Fix return type
+    def _format_timeframe(self, timeframe: TimeFrames, delta: Union[float, int]) -> Tuple[str, str]:  # type: ignore
         return (
             self.timeframes[timeframe][0].format(abs(delta)),
             self.timeframes[timeframe][1].format(abs(delta)),
         )
```
isac322, line 928: The return type of this particular method is not match with others. Should I change return type of `_format_timeframe()` and signature of all related methods?

arrow/locales.py
```diff
@@ -1121,19 +1197,25 @@ class KoreanLocale(Locale):
...
+    ) -> str:
         if timeframe in ("day", "days"):
-            special = self.special_dayframes.get(delta)
+            # TODO: comment on github PR
+            special = self.special_dayframes.get(int(delta))
             if special:
                 return special
         elif timeframe in ("year", "years"):
-            special = self.special_yearframes.get(delta)
+            special = self.special_yearframes.get(int(delta))
```
isac322, line 1218: `delta` could be `float` but `special_dayframes` is `Mapping[int, str]`

arrow/locales.py
```diff
@@ -1212,14 +1294,14 @@ class DutchLocale(Locale):
     day_abbreviations = ["", "ma", "di", "wo", "do", "vr", "za", "zo"]
 
 
-class SlavicBaseLocale(Locale):
-    def _format_timeframe(self, timeframe, delta):
+class SlavicBaseLocale(Locale, metaclass=ABCMeta):
```
isac322, line 1297: Added `ABCMeta`

arrow/locales.py
```diff
@@ -1212,14 +1294,14 @@ class DutchLocale(Locale):
...
+    timeframes: ClassVar[Mapping[TimeFrames, Union[str, Sequence[str]]]]
 
+    def _format_timeframe(self, timeframe: TimeFrames, delta: Union[float, int]) -> str:
         form = self.timeframes[timeframe]
         delta = abs(delta)
 
-        if isinstance(form, list):
-
+        if isinstance(form, Sequence):
```
isac322, line 1304: Can I modify like this? It feels more robust to me.

arrow/locales.py
```diff
@@ -1717,7 +1799,7 @@ class MacedonianLocale(SlavicBaseLocale):
     ]
 
 
-class GermanBaseLocale(Locale):
+class GermanBaseLocale(Locale, metaclass=ABCMeta):
```
isac322, line 1802: Added `ABCMeta`

arrow/locales.py
```diff
@@ -2566,22 +2653,22 @@ class MoroccoArabicLocale(ArabicLocale):
...
+    def _format_timeframe(self, timeframe: TimeFrames, delta: Union[float, int]) -> str:
+        form = self.timeframes[timeframe]
+        if isinstance(form, Sequence):
+            if delta < 0:
+                form = form[0]
+            elif delta > 0:
+                form = form[1]
 
-        return timeframe.format(abs(delta))
+        return form.format(abs(delta))
```
isac322, line 2664: Refactoring.
The type of value at `timeframes` is `Union[Tuple[str, str], str]]`

arrow/locales.py
```diff
@@ -2921,32 +3010,40 @@ class CzechLocale(Locale):
...
+        if isinstance(form, Sequence):
+            if 2 <= abs_delta % 10 <= 4 and (
+                abs_delta % 100 < 10 or abs_delta % 100 >= 20
+            ):
                 form = form[0]
             else:
                 form = form[1]
 
-        return form.format(delta)
+        return form.format(abs_delta)
```
isac322, line 3038: Refactoring for better type analysis. Same functionality (I think)

arrow/locales.py
```diff
@@ -3011,26 +3108,32 @@ class SlovakLocale(Locale):
...
```
isac322, line 3136: Refactoring for better type analysis. Same functionality (I think)

arrow/parser.py
```diff
@@ -339,6 +507,7 @@ def _parse_token(self, token, value, parts):
             parts["year"] = 1900 + value if value > 68 else 2000 + value
 
         elif token in ["MMMM", "MMM"]:
+            # FIXME: month_number() is nullable
             parts["month"] = self.locale.month_number(value.lower())
```
isac322, line 511: We should refactor this.
`month_number()` is nullable

jadchaar (commented):
Small comments on first glance of a few files.

arrow/locales.py
jadchaar: I don't think so. It seems like `Sequence` encompasses more than `list` according to this: https://docs.python.org/3/glossary.html

Quote from the glossary: "Some built-in sequence types are list, str, tuple, and bytes"

arrow/locales.py
jadchaar: Mind explaining the purpose of a metaclass?

arrow/locales.py
jadchaar: Let's make all of the `_format_timeframe` functions consistent. This might be related to https://github.com/arrow-py/arrow/issues/789.

arrow/factory.py
```diff
@@ -26,10 +27,67 @@ class ArrowFactory:
 
     """
 
-    def __init__(self, type=Arrow):
+    type: Type[Arrow]
```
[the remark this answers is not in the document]
isac322: Just a typing for [instance variable](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes)

arrow/locales.py
isac322: Can't specific reason... 😅
I'll remove it and discuss in another PR

arrow/arrow.py
isac322: rewrote!

arrow/locales.py
```diff
@@ -2566,22 +2651,23 @@ class MoroccoArabicLocale(ArabicLocale):
...
+        form = self.timeframes[timeframe]
+        if isinstance(form, tuple):
+            if delta < 0:
+                form = form[0]
+            elif delta > 0:
+                form = form[1]
+            # FIXME: handle when delta is 0
 
-        return timeframe.format(abs(delta))
+        return form.format(abs(delta))  # type: ignore
```
isac322, line 2663: We should handle when delta is 0. Otherwise mypy `form` in return line can be `Tuple`

arrow/formatter.py
isac322: Please review refactored code!

arrow/locales.py
isac322: Got it! reverted it.

arrow/arrow.py
```diff
@@ -1390,84 +1496,140 @@ def for_json(self):
...
-    def __add__(self, other):
+    @overload
+    def __add__(self, other: timedelta) -> "Arrow":
+        pass  # pragma: no cover
+
+    @overload
+    def __add__(self, other: relativedelta) -> "Arrow":
+        pass  # pragma: no cover
+
+    def __add__(self, other: Any) -> "Arrow":
```
jadchaar, line 1507: We need overloads here since we need to trigger a `NotImplemented` error in the case of an inappropriate type correct?

arrow/arrow.py
```diff
@@ -1524,7 +1688,9 @@ def _get_frames(cls, name):
         )
 
     @classmethod
-    def _get_iteration_params(cls, end, limit):
+    def _get_iteration_params(
+        cls, end: Any, limit: Union[int, None]
```
jadchaar, line 1692: `Union[int, None]` should be changed to `Optional[int]`.

Think it's worth refactoring this function to use `kwargs` or `args`, and adding two overloads, one with the end argument, and one with the limit argument? That way we enforce the fact that this should only take one or the other?

jadchaar (changes requested):
Initial review up to arrow.py. I will continue my review over the next few days!

arrow/arrow.py
```diff
@@ -50,55 +111,77 @@ class Arrow:
...
+        "second": 1.0,
+        "minute": _SECS_PER_MINUTE,
+        "hour": _SECS_PER_HOUR,
+        "day": _SECS_PER_DAY,
+        "week": _SECS_PER_WEEK,
+        "month": _SECS_PER_MONTH,
+        "year": _SECS_PER_YEAR,
+    }
+
+    _datetime: dt_datetime
```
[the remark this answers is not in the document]
isac322: This isn't Class variable. It's a [Instance Variable](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes)

arrow/arrow.py
```diff
@@ -50,55 +111,77 @@ class Arrow:
...
-    _MONTHS_PER_QUARTER = 3
-    _SECS_PER_MINUTE = float(60)
-    _SECS_PER_HOUR = float(60 * 60)
-    _SECS_PER_DAY = float(60 * 60 * 24)
-    _SECS_PER_WEEK = float(60 * 60 * 24 * 7)
-    _SECS_PER_MONTH = float(60 * 60 * 24 * 30.5)
-    _SECS_PER_YEAR = float(60 * 60 * 24 * 365)
+    resolution: ClassVar[timedelta] = dt_datetime.resolution
+    min: ClassVar["Arrow"]
+    max: ClassVar["Arrow"]
```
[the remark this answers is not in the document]
isac322: We can't I think. Python does not allow to use `self` within class declaration context and to reference class itself.

arrow/arrow.py
```diff
@@ -50,55 +111,77 @@ class Arrow:
...
+    _ATTRS_PLURAL: Final[List[str]] = [f"{a}s" for a in _ATTRS]
+    _MONTHS_PER_QUARTER: Final[int] = 3
+    _SECS_PER_MINUTE: Final[float] = float(60)
+    _SECS_PER_HOUR: Final[float] = float(60 * 60)
+    _SECS_PER_DAY: Final[float] = float(60 * 60 * 24)
+    _SECS_PER_WEEK: Final[float] = float(60 * 60 * 24 * 7)
+    _SECS_PER_MONTH: Final[float] = float(60 * 60 * 24 * 30.5)
+    _SECS_PER_YEAR: Final[float] = float(60 * 60 * 24 * 365)
+
+    _SECS_MAP: Final[Mapping[TimeFrames, float]] = {
```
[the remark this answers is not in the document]
isac322: Yeap. `Mapping` does not enforce to declare all possible values as key. Just like `Mapping[str, in] = dict(a=1)` is OK.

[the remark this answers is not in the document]
isac322: If we have to enforce this dict to have certain set, then we have to use [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict)l.

arrow/arrow.py
```diff
@@ -7,14 +7,75 @@
 
 import calendar
 import sys
-from datetime import datetime, timedelta
+from datetime import date
+from datetime import datetime as dt_datetime
+from datetime import time as dt_time
```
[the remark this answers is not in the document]
isac322: There are `Arrow.datetime()` and `Arrow.time()`. So I think this aliasing remove confusing.

arrow/arrow.py
```diff
@@ -155,7 +238,11 @@ def utcnow(cls):
         )
 
     @classmethod
-    def fromtimestamp(cls, timestamp, tzinfo=None):
+    def fromtimestamp(
+        cls,
+        timestamp: Union[SupportsFloat, str],
```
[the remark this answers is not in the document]
isac322: If we declare `SupportsFloat`, we can accept all class that implemented `__float__()`.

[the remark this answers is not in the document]
isac322: Ah.. then I also have to modify `util.is_timestamp()` too.

arrow/arrow.py
```diff
@@ -614,15 +730,15 @@ def __getattr__(self, name):
...
 
             if value is not None:
                 return value
 
-        return object.__getattribute__(self, name)
+        return cast(int, object.__getattribute__(self, name))
 
     @property
-    def tzinfo(self):
+    def tzinfo(self) -> Optional[dt_tzinfo]:
```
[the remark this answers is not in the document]
isac322: We have `tzinfo` setter and it accepts `None` as tzinfo. So we can not assure that this method always returns non null value.

arrow/arrow.py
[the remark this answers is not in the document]
isac322: I think using `SupportsFloat` is good but also it feels like we have to discuss on another PR. So in this PR I will use `Union[int, float, str]`.

arrow/arrow.py
```diff
@@ -975,7 +1093,7 @@ def humanize(
         if isinstance(granularity, list) and len(granularity) == 1:
             granularity = granularity[0]
 
-        delta = int(round(util.total_seconds(self._datetime - dt)))
+        delta: Union[int, float] = int(round(util.total_seconds(self._datetime - dt)))
```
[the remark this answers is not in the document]
isac322: It's because of later usage of the variable. but I think I can refactor it.

arrow/arrow.py
isac322: Yeap. if I remove overloads and set `other` as `Union[timedelta, relativedelta]`, mypy blame that `return NotImplemented` line is unreachable.

arrow/arrow.py
isac322: Yeap we can achieve enforcing without using `kwargs` or `args`. but is it worth to refactor? I saw only one usage and even there the parameters are not manipulated value. They just pass from their arguments.

arrow/arrow.py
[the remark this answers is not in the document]
isac322: We can restrict the setter also to prohibit `None`, but it will break backward compatibility.

arrow/arrow.py
```diff
@@ -791,11 +905,11 @@ def replace(self, **kwargs):
 
         # TODO revisit this once we drop support for 2.7/3.5
         if fold is not None:
-            current = dateutil_tz.enfold(current, fold=fold)
+            current = dateutil_tz.enfold(current, fold=fold)  # type: ignore
 
         return self.fromdatetime(current)
 
-    def shift(self, **kwargs):
+    def shift(self, **kwargs: Any) -> "Arrow":
```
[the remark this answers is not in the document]
isac322: I don't think overloading is fit for this. If we use overload, we have to declare all parameter combinations. Instead I suggest to  use [Keyword Only Arguments](https://www.python.org/dev/peps/pep-3102/). but we have to set default value for all possible arguments (e.g. `None` if no shift) and add (default value) filtering logic on first.

arrow/locales.py
isac322: I think the issue can handle in separate PR. Can I handle on another PR?

isac322: What you mean is we have to change return type of all `_format_timeframe()`? Then we also have to modify logic of `describe()` and `describe_multi()`. but I think I can handle it because I'm not that aware of locale things.. Can you help this? @jadchaar

arrow/arrow.py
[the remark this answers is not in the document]
jadchaar: I am pretty sure that variables declared outside of a function and within the class body are static/class variables: https://stackoverflow.com/questions/68645/are-static-class-variables-possible-in-python.

arrow/arrow.py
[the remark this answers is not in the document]
jadchaar: Ah yeah that is a good point. Like you said, we can do it simple for now, but make an issue and tackle the support for any float-like object in a separate PR.

arrow/arrow.py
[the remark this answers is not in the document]
jadchaar: Yeah datetime allows for you to pass None, but I do not think Arrow itself ever sets tzinfo to None internally since we disallow naive datetime formation. I am pretty that in all cases where None is passed in, we substitute in `dateutil.tz.tzutc()` to prevent the creation of naive datetimes. CC @systemcatch to confirm this, but I am pretty confident on that.

arrow/arrow.py
jadchaar: Yeah it might not be worth the refactor. We have been looking to refactor span/interval for a while now, and when we do, I bet this function will be nuked anyways.

arrow/arrow.py
```diff
@@ -1032,81 +1148,69 @@ def humanize(
...
+                        value = sign * _delta / self._SECS_MAP[_frame]
+                        _delta %= self._SECS_MAP[_frame]
+                        if trunc(abs(value)) != 1:
+                            timeframes.append((cast(TimeFrames, _frame + "s"), value))
+                        else:
+                            timeframes.append((_frame, value))
+                    return _delta
+
+                delta = float(delta_second)
+                frames: Tuple[TimeFrames, ...] = (
```
jadchaar, line 1196: Does the `...` syntax here mean we are only using a subset of `TimeFrames`? Sorry I am new to mypy/typing syntax hah!

jadchaar: Oh nevermind... this seems to just be spreading the same type over the length of the tuple!

arrow/arrow.py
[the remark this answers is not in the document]
systemcatch: Hey @jadchaar, interestingly the `tzinfo` setter will cause a `None` value for tzinfo to occur.

```shell
Python 3.8.3 (default, Jul  7 2020, 18:57:36) 
[GCC 9.3.0] on linux
>>> import arrow
>>> dt=arrow.utcnow()
>>> dt
<Arrow [2020-11-24T20:23:04.371964+00:00]>
>>> dt.tzinfo
tzutc()
>>> dt.tzinfo=None
>>> dt.tzinfo
>>> dt
<Arrow [2020-11-24T20:23:04.371964]>
```

I think we should prohibit this to avoid weird results. The best approach would be to remove the `tzinfo` setter entirely because that also helps us with `Arrow` objects being hashable yet also mutable.

arrow/formatter.py
```diff
@@ -116,7 +130,11 @@ def _format_token(self, dt, token):
         if token in ["ZZ", "Z"]:
             separator = ":" if token == "ZZ" else ""
             tz = dateutil_tz.tzutc() if dt.tzinfo is None else dt.tzinfo
-            total_minutes = int(util.total_seconds(tz.utcoffset(dt)) / 60)
+            utcoffset = tz.utcoffset(dt)
+            if utcoffset is None:
+                return None
```
jadchaar, line 135: utcoffset should never be none in Arrow since we will always have aware datetime objects (never naive). According to the docs (https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive), `utc.offset` will never return None if a datetime object is aware. So I do not think we need to worry about utcoffset ever being None in our case. CC @systemcatch to hear his thoughts.

systemcatch: I don't think we can end up with utcoffset being None here given the line above.

jadchaar (commented):
Slowly working through locales, but here are some comments I had in the mean time.

arrow/arrow.py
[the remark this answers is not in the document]
isac322: If there are initial value, then the it becomes Class variable. but because I only declare variable without initialization, it is just type declaration for Instance variable. You can check [PEP-526](https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations)
```python
Python 3.8.6 (default, Sep 30 2020, 04:00:38) 
[GCC 10.2.0] on linux
>>> import arrow
>>> arrow.Arrow._datetime
AttributeError: type object 'Arrow' has no attribute '_datetime'
>>> arrow.Arrow._SECS_MAP
{'second': 1.0, 'minute': 60.0, 'hour': 3600.0, 'day': 86400.0, 'week': 604800.0, 'month': 2635200.0, 'year': 31536000.0}
```

arrow/arrow.py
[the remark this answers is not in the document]
isac322: That's my point! Also I'm agree with removing `tzinfo` setter. `Arrow.replace()` have same functionality with `None` value filtering.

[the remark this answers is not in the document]
isac322: But I think removing has to be discussed in another PR or issue.

arrow/formatter.py
isac322: `Arrow` passes `self._datetime` to this method and it can be naive as [discuss above](https://github.com/arrow-py/arrow/pull/883#discussion_r529185727).
And I also think this method should handle naive `datetime`, because `DateTimeFormatter.format` is public anybody can pass any `datetime` object.

systemcatch: Yes `datetime` can be naive (None) but in that case we end up with the following.

```shell
Python 3.8.3 (default, Jul  7 2020, 18:57:36) 
[GCC 9.3.0] on linux
>>> from dateutil import tz
>>> tutc=tz.tzutc()
>>> dt=datetime(2019, 5, 6, 7)
>>> utcoffset=tutc.utcoffset(dt)
>>> utcoffset
datetime.timedelta(0)
```

isac322: @systemcatch 
I saw [Python datetime doc](https://docs.python.org/3/library/datetime.html#determining-if-an-object-is-aware-or-naive) and it saids `d.tzinfo.utcoffset(d) does not return None`. So I created a [test case](https://github.com/arrow-py/arrow/pull/883/files#diff-e0cd3f215f68d0608393c012e268cb2f896c1e88315c97fa77e114d400f5283bR134) for the scenario. please check it.

```
Python 3.8.6 (default, Sep 30 2020, 04:00:38) 
[GCC 10.2.0] on linux
>>> from datetime import tzinfo, datetime
>>> class IncompleteTzinfo(tzinfo):
...   def utcoffset(self, dt):
...     return None
... 
>>> dt = datetime.utcnow()
>>> tz = IncompleteTzinfo()
>>> tz.utcoffset(dt)
>>> 

```

systemcatch: Interesting, someone would need to make a subclass specifically to do this right?

I suppose this adds extra safety with no downside so let's leave it in.

isac322: OK then I will remove this handling codes and just cast it to `timedelta`

nkitsaini:
I am excited for type hint in arrow. Is there anything I can do to help this get merged (faster) in master ?

jadchaar:
Hi @nkitsaini i have been a bit busy lately, but I should be more free to finish reviewing this in the next couple of weeks. We are hoping to ship this as part of v1.0 by the end of the year!

nkitsaini:
Thanks for quick reply (and sorry for my late reply). No worries actually, end of the year is far better then what I was expecting.

jadchaar:
Hey @isac322 I should be able to review this in the next couple of weeks as things wind down for me. Mind getting this branch up to date with master?

arrow/locales.py
```diff
@@ -188,24 +245,31 @@ def meridian(self, hour, token):
...
 
-    def _name_to_ordinal(self, lst):
+    def _name_to_ordinal(self, lst: Sequence[str]) -> Dict[str, int]:
         return dict(map(lambda i: (i[1].lower(), i[0] + 1), enumerate(lst[1:])))
 
-    def _format_timeframe(self, timeframe, delta):
-        return self.timeframes[timeframe].format(trunc(abs(delta)))
+    def _format_timeframe(self, timeframe: TimeFrames, delta: Union[float, int]) -> str:
+        # TODO: remove cast
+        return cast(str, self.timeframes[timeframe]).format(trunc(abs(delta)))
```
systemcatch, line 265: Hey @isac322 do you want to remove this?

arrow/locales.py
jadchaar: I agree with the change--I don't think mixing tuples and lists in a container will see much usage.

arrow/locales.py
```diff
@@ -92,7 +145,11 @@ def describe(self, timeframe, delta=0, only_distance=False):
 
         return humanized
 
-    def describe_multi(self, timeframes, only_distance=False):
+    def describe_multi(
+        self,
+        timeframes: Collection[Tuple[TimeFrames, Union[int, float]]],
```
jadchaar, line 150: Are we using a collection here to encompass both the possibility of both `tuple` and `list`? Do you think it's problematic since collections span beyond these two types?

jadchaar (changes requested):
Hi @isac322, thanks again for the great work. I did another pass on the files, and had a few remaining bits of feedback and questions!

Happy holidays!

arrow/formatter.py
```diff
@@ -22,19 +31,24 @@ class DateTimeFormatter:
...
+
+    def __init__(self, locale: str = "en_us") -> None:
 
         self.locale = locales.get_locale(locale)
 
-    def format(cls, dt, fmt):
+    def format(cls, dt: datetime, fmt: str) -> str:
 
-        return cls._FORMAT_RE.sub(lambda m: cls._format_token(dt, m.group(0)), fmt)
+        # FIXME: _format_token() is nullable
```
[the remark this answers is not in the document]
isac322: @jadchaar Nop. `DateTimeFormatter._format_token()` still can returns `None`. (if `token` did not match with all conditions) But because `DateTimeFormatter._format_token()` is internal purpose, we can override return type as `str` and remove cast and FIXME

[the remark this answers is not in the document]
isac322: @jadchaar And plus, we have to make sure that `dt` is not naive. `dt.tzname()` in `DateTimeFormatter._format_token()` can return `None` if `dt` is naive.

arrow/parser.py
```diff
@@ -218,7 +308,9 @@ def parse(self, datetime_string, fmt, normalize_whitespace=False):
             return self._parse_multiformat(datetime_string, fmt)
 
         try:
-            fmt_tokens, fmt_pattern_re = self._generate_pattern_re(fmt)
+            fmt_tokens, fmt_pattern_re = self._generate_pattern_re(
+                fmt
+            )  # type: List[_FORMAT_TYPE], Pattern[str]
```
jadchaar, line 313: This can be changed to a type annotation right before the unpacking:

```python
fmt_tokens: List[_FORMAT_TYPE]
fmt_pattern_re: Pattern[str]
fmt_tokens, fmt_pattern_re = self._generate_pattern_re(
    fmt
)
```

arrow/parser.py
```diff
@@ -231,8 +323,9 @@ def parse(self, datetime_string, fmt, normalize_whitespace=False):
                 f"Failed to match '{fmt}' when parsing '{datetime_string}'."
             )
 
-        parts = {}
-        for token in fmt_tokens:
+        parts: _Parts = {}
+        for token in fmt_tokens:  # type: _FORMAT_TYPE
```
jadchaar, line 327: If we change the declaration of `fmt_tokens` above, this comment can probably be removed.

arrow/parser.py
```diff
@@ -327,7 +425,76 @@ def _generate_pattern_re(self, fmt):
...
+        parts: _Parts,
+    ) -> None:
+        ...  # pragma: no cover
+
+    def _parse_token(
+        self,
+        token: Any,
+        value: Any,
+        parts: Any,
+    ) -> None:
```
jadchaar, line 497: Mind explaining the rationale for `value`'s type in these overloads?

From this:
https://github.com/arrow-py/arrow/blob/ea9f325dea89194fbd07cf74ad34a7e95f65ae4f/arrow/parser.py#L235-L248

It seems like the expected types for `value` should be `value: Union[Tuple[str, str, str], str]`

arrow/parser.py
jadchaar: Would you like to take on this refactor?

arrow/parser.py
```diff
@@ -582,7 +753,13 @@ def parse(cls, tzinfo_string):
...
 
             if iso_match:
-                sign, hours, minutes = iso_match.groups()
+                (
+                    sign,
+                    hours,
+                    minutes,
+                ) = (
+                    iso_match.groups()
+                )  # type: Optional[str], str, Union[str, int, None]
```
jadchaar, line 762: Should we declare `sign`, `hours`, and `minutes` types before to avoid the commented types?

jadchaar (changes requested):
Slowly working through all the files. I have 2 more left, but this is another set of feedback @isac322.

Thanks for the continued hard work on this PR! @systemcatch and I are looking to get it merged soon :).

arrow/arrow.py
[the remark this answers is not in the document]
jadchaar: Ah yes, fantastic!

arrow/arrow.py
```diff
@@ -213,7 +299,9 @@ def utcfromtimestamp(cls, timestamp):
         )
 
     @classmethod
-    def fromdatetime(cls, dt, tzinfo=None):
+    def fromdatetime(
+        cls, dt: Union[dt_datetime, "Arrow"], tzinfo: Optional[TZ_EXPR] = None
```
jadchaar, line 303: `dt` should be of type `dt_datetime` only. This cannot be used to construct an object from an existing Arrow object.

arrow/arrow.py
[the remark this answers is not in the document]
jadchaar: The setter is going to be removed in https://github.com/arrow-py/arrow/pull/902. Arrow objects should be immutable to prevent messing with hash behavior.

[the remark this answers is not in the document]
jadchaar: Once the setter is removed, we should be able to change the return type from `Optional[dt_tzinfo]` to `dt_tzinfo` since Arrow objects are never naive.

arrow/arrow.py
```diff
@@ -1445,84 +1558,140 @@ def for_json(self):
...
 
-    def __rsub__(self, other):
+    @overload
+    def __rsub__(self, other: dt_datetime) -> timedelta:
+        pass  # pragma: no cover
 
-        if isinstance(other, datetime):
+    @overload
+    def __rsub__(self, other: Any) -> timedelta:
+        pass  # pragma: no cover
```
jadchaar, line 1606: What's the purpose of this overload? Isn't it identical to the original `__rsub__` definition below?

arrow/arrow.py
```diff
@@ -1445,84 +1558,140 @@ def for_json(self):
...
         return not self.__eq__(other)
 
-    def __gt__(self, other):
+    @overload
+    def __gt__(self, other: "Arrow") -> bool:
+        pass  # pragma: no cover
+
+    @overload
+    def __gt__(self, other: dt_datetime) -> bool:
+        pass  # pragma: no cover
```
jadchaar, line 1637: Can these overloads be combined:

```python
    @overload
    def __gt__(self, other: Union[dt_datetime, "Arrow"]) -> bool:
        pass  # pragma: no cover
```

arrow/arrow.py
```diff
@@ -1445,84 +1558,140 @@ def for_json(self):
...
         return self._datetime >= self._get_datetime(other)
 
-    def __lt__(self, other):
+    @overload
+    def __lt__(self, other: "Arrow") -> bool:
+        pass  # pragma: no cover
+
+    @overload
+    def __lt__(self, other: dt_datetime) -> bool:
+        pass  # pragma: no cover
```
jadchaar, line 1667: Same here. Can these overloads be combined using a `Union` for the `other` type?

arrow/arrow.py
```diff
@@ -1445,84 +1558,140 @@ def for_json(self):
...
         return self._datetime < self._get_datetime(other)
 
-    def __le__(self, other):
+    @overload
+    def __le__(self, other: "Arrow") -> bool:
+        pass  # pragma: no cover
+
+    @overload
+    def __le__(self, other: dt_datetime) -> bool:
+        pass  # pragma: no cover
```
jadchaar, line 1682: Same here. Do let me know if these can or cannot be combined :)!

jadchaar (changes requested):
I was finally able to get through all the files and left a final round of feedback.

Happy holidays to you @isac322 and I hope you can address the comments after the holidays or whenever you get a free moment :).

slavaatsig:
Why not support [PEP-563](https://www.python.org/dev/peps/pep-0563/)?

jadchaar:
> Why not support [PEP-563](https://www.python.org/dev/peps/pep-0563/)?

Doesn't PEP-563 only support Python 3.7+? We must support 3.6+.

jadchaar:
Hey @isac322, any progress on addressing feedback? We really appreciate your work here, and would love to get this merged soon! This is the last big puzzle piece for version 1.0 :).

isac322:
@jadchaar Sorry for late reply. I'm currently quite busy for my marriage and company reorganization. Can I check this on last week of January?

jadchaar:
> @jadchaar Sorry for late reply. I'm currently quite busy for my marriage and company reorganization. Can I check this on last week of January?

No problem at all buddy, take your time! Also, congratulations and best of luck with the marriage/reorganization :).

arrow/locales.py
isac322: @jadchaar  If I change return type of `_format_timeframe` to `Union[str, Tuple[str, str]]`, mypy returns

```
arrow/locales.py:143: error: Argument 1 to "_format_relative" of "Locale" has incompatible type "Union[Tuple[str, str], str]"; expected "str"
arrow/locales.py:145: error: Incompatible return value type (got "Union[Tuple[str, str], str]", expected "str")
arrow/locales.py:163: error: Argument 1 to "join" of "str" has incompatible type "List[Union[Tuple[str, str], str]]"; expected "Iterable[str]"
```

I think just add `type: ignore` on this PR and refactor locale related code using strategy pattern or something in another PR. The refactoring must be quite big I think.

arrow/parser.py
isac322: @jadchaar When I change type of `parser._Parts.month` from `int` to `Optional[int]`, mypy prints

```
arrow/parser.py:656: error: Argument 2 to "datetime" has incompatible type "Optional[int]"; expected "int"
arrow/parser.py:697: error: Argument "month" to "datetime" has incompatible type "Optional[int]"; expected "int"
```

But I don't have any clue to fix this because I'm not familiar with `_build_datetime()` and `parts` is mutable. Can you help with this?

arrow/arrow.py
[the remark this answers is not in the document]
isac322: @jadchaar @systemcatch Changed it!

arrow/locales.py
```diff
@@ -1,9 +1,55 @@
...
+    "years",
+    "2-hours",
+    "2-days",
+    "2-weeks",
+    "2-months",
+    "2-years",
+]
+
+_TimeFrameElements = Union[
+    str, Sequence[str], Mapping[str, str], Mapping[str, Sequence[str]]
```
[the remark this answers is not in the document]
isac322: @jadchaar  I think most of usage of `_TimeFrameElements` can handle regardless of container type. If I declare like this, we can accept `tuple` as `timeframes`.

arrow/locales.py
```diff
@@ -188,24 +245,31 @@ def meridian(self, hour, token):
...
         :param n: an integer
         """
         return self._ordinal_number(n)
 
-    def _ordinal_number(self, n):
+    def _ordinal_number(self, n: int) -> str:
         return f"{n}"
 
-    def _name_to_ordinal(self, lst):
+    def _name_to_ordinal(self, lst: Sequence[str]) -> Dict[str, int]:
```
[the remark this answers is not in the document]
isac322: Same here. `_name_to_ordinal(("", "Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"))` is perfectly OK.

arrow/locales.py
isac322: @systemcatch Yeap, I want it because `cast` is some kind of monkey patch. But I think just leave it on this PR and refactor whole Locale part later is better approach.

arrow/formatter.py
```diff
@@ -116,7 +130,8 @@ def _format_token(self, dt, token):
         if token in ["ZZ", "Z"]:
             separator = ":" if token == "ZZ" else ""
             tz = dateutil_tz.tzutc() if dt.tzinfo is None else dt.tzinfo
-            total_minutes = int(tz.utcoffset(dt).total_seconds() / 60)
+            # https://github.com/arrow-py/arrow/pull/883#discussion_r529866834
```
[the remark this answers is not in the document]
isac322: Done!

arrow/util.py
```diff
@@ -1,11 +1,14 @@
...
+from typing import Any, Optional, cast
 
 from dateutil.rrule import WEEKLY, rrule
 
 from arrow.constants import MAX_TIMESTAMP, MAX_TIMESTAMP_MS, MAX_TIMESTAMP_US
 
 
-def next_weekday(start_date, weekday):
+def next_weekday(
+    start_date: Optional[datetime.date], weekday: int
```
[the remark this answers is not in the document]
isac322: @jadchaar But [`dateutil.rrule`](https://github.com/python/typeshed/blob/8f7e2216b66b9e8181b119fa2f7d560fca987605/stubs/python-dateutil/dateutil/rrule.pyi#L39) can handle those cases.

arrow/locales.py
```diff
@@ -1,9 +1,55 @@
...
+    Union,
+    cast,
+)
+
+if sys.version_info < (3, 8):  # pragma: no cover
+    from typing_extensions import Literal
+else:
+    from typing import Literal  # pragma: no cover
+
+TimeFrames = Literal[
```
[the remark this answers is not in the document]
isac322: Done!

arrow/locales.py
isac322: Fixed it to `Sequence` because after merging master, there are usage of random access on `timeframes`.

arrow/constants.py
```diff
@@ -28,3 +28,6 @@
 MAX_TIMESTAMP: Final[float] = _MAX_TIMESTAMP
 MAX_TIMESTAMP_MS: Final[float] = MAX_TIMESTAMP * 1e3
 MAX_TIMESTAMP_US: Final[float] = MAX_TIMESTAMP * 1e6
+
+MAX_ORDINAL = datetime.max.toordinal()
+MIN_ORDINAL = 1
```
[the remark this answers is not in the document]
isac322: Done!

arrow/parser.py
isac322: Done!

arrow/parser.py
isac322: Done!

arrow/parser.py
systemcatch: @isac322 maybe we can refactor this after the PR is merged?

systemcatch (approved):
Hey @isac322 thanks for coming back and implementing the suggested changes. It's a big piece of work.

I think this PR is ready to be merged and any minor issues can be fixed afterwards.

@jadchaar , @krisfremen time to merge this?

arrow/locales.py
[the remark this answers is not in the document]
isac322: Done!

arrow/parser.py
isac322: Done!

arrow/locales.py
```diff
@@ -2566,22 +2651,22 @@ class MoroccoArabicLocale(ArabicLocale):
...
         if delta < 0:
-            timeframe = timeframe[0]
+            form = form[0]
         elif delta > 0:
-            timeframe = timeframe[1]
+            form = form[1]
+            # FIXME: handle when delta is 0
 
-        return timeframe.format(abs(delta))
+        return form.format(abs(delta))  # type: ignore
```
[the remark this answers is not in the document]
isac322: Yeap, [This issue](https://github.com/arrow-py/arrow/pull/883#discussion_r525832604) should must be handled in advance.

arrow/arrow.py
isac322: Done! Thanks for catching

arrow/arrow.py
isac322: Originally I want to express `NotImplemented` path in type annotation. but currently python typing does not support this feature. (https://github.com/python/mypy/issues/4791) And I found [related work](https://github.com/pantsbuild/pants/pull/11506#discussion_r567252681) so I remove all `overload`s except `__sub__`.

arrow/arrow.py
isac322: same here. https://github.com/arrow-py/arrow/pull/883#discussion_r575660242

arrow/arrow.py
isac322: same here. https://github.com/arrow-py/arrow/pull/883#discussion_r575660242

arrow/arrow.py
isac322: same here. https://github.com/arrow-py/arrow/pull/883#discussion_r575660242

arrow/locales.py
```diff
@@ -60,25 +106,32 @@ class Locale:
         "years": "",
     }
 
-    meridians = {"am": "", "pm": "", "AM": "", "PM": ""}
+    meridians: ClassVar[Dict[str, str]] = {"am": "", "pm": "", "AM": "", "PM": ""}
+
+    past: ClassVar[str]
+    future: ClassVar[str]
```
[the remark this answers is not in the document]
isac322: If we have to have initial value for those variables, then I have to declare `past` and `future` as `ClassVar[Optional[str]]`.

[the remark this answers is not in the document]
isac322: But because I have to refactor locale part anyway, I just set `ClassVar[Optional[str]]` on this PR.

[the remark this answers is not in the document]
isac322: And If I do that, mypy complains on `Locale._format_relative()`
```
arrow/locales.py:279: error: Item "None" of "Optional[str]" has no attribute "format"
```

isac322:
@jadchaar @systemcatch Could you please mark conversation as resolved if you satisfied on changes for future work? Because there are many conversation, it's quite hard to distinguish TODOs 😢.

arrow/parser.py
isac322: @jadchaar  Ah, when I add type annotation on existing code, I do not consider caller side's input and always try to consider inside of the function. because it helps me to narrow the signature down.
But it could be too much for internal purpose method. Do you want me to simplify this?

systemcatch:
I've gone through and marked many conversations as resolved, the few remaining ones need Jad's input.

jadchaar:
> I've gone through and marked many conversations as resolved, the few remaining ones need Jad's input.

Fantastic and thanks for the hard work @isac322, we really do appreciate it. I will give this a good final look either today or tomorrow!

arrow/formatter.py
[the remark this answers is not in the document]
jadchaar: @isac322 is this FIXME good to be removed?

arrow/formatter.py
```diff
@@ -22,19 +31,24 @@ class DateTimeFormatter:
...
+    def format(cls, dt: datetime, fmt: str) -> str:
 
-        return cls._FORMAT_RE.sub(lambda m: cls._format_token(dt, m.group(0)), fmt)
+        # FIXME: _format_token() is nullable
+        return cls._FORMAT_RE.sub(
+            lambda m: cast(str, cls._format_token(dt, m.group(0))), fmt
+        )
 
-    def _format_token(self, dt, token):
+    def _format_token(self, dt: datetime, token: Optional[str]) -> Optional[str]:
```
jadchaar, line 51: In what case can `_format_token` return `None`?

arrow/parser.py
```diff
@@ -337,7 +504,8 @@ def _parse_token(self, token, value, parts):
             parts["year"] = 1900 + value if value > 68 else 2000 + value
 
         elif token in ["MMMM", "MMM"]:
-            parts["month"] = self.locale.month_number(value.lower())
+            # FIXME: month_number() is nullable
+            parts["month"] = self.locale.month_number(value.lower())  # type: ignore
```
jadchaar, line 508: @isac322 Is this FIXME good to be removed?

jadchaar (approved):
LGTM, I had a few final comments though. 

Thanks for your patience @isac322! I have been trying to integrate typing into my own projects recently, but your typing skillset is much more advanced so it took some time for me to understand it :).

arrow/formatter.py
isac322: @jadchaar  If token is `None` or any `str` that does not match with if conditions.

arrow/formatter.py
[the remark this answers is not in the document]
isac322: @jadchaar I can't because return type of `_format_token()` is still `Optional[str]` as explained [comment below](https://github.com/arrow-py/arrow/pull/883#discussion_r575953514)

arrow/parser.py
isac322: @jadchaar  Hmmmm. I'm not sure `Locale.month_number()` always return `str`.

isac322:
@jadchaar @systemcatch  It was a long journey, but it was fun. Thanks for your work!

arrow/formatter.py
jadchaar: Could these FIXMEs, if left unresolved, cause issues for users in regular use? When will the nullable issues you mark with the FIXME label cause problems?

isac322: It could be. but because I didn't change actual behavior of `formatter.py`, this PR will not introduce new error to `formatter.py`.
BTW, I can make `token` as `str` because [`Match.group()`](https://docs.python.org/3/library/re.html#re.Match.group) always return `AnyStr`.

isac322: If we can assure that `token` always match one branch in `_format_token`, I have to narrow down type of `token` to some `Literal`.

jadchaar: To test things out, I put a `raise ValueError` at the bottom of the control statements to see if I could possibly get anything to fall through, but I was unable to ever reach beyond the conditionals.

I tried these:

```python
now = arrow.now()

now.format()
now.format("")
now.format(None)
```

The only thing that caused trouble was the final None one, but it failed on the call to `sub` before even going into `_format_token`. I cannot seem to find a case that digs all the way down to the bottom of the `_format_token` function, so we might be safe to say it always returns a string and never `None`.

jadchaar: When I run the code with the `raise ValueError` at the end of `_format_token`, the only test that fails in the test suite is `test_nonsense` in `test_formatter.py`:


```console
====================================================================================================================================================== FAILURES ======================================================================================================================================================
_______________________________________________________________________________________________________________________________________ TestFormatterFormatToken.test_nonsense _______________________________________________________________________________________________________________________________________

self = <tests.test_formatter.TestFormatterFormatToken object at 0x1034c1d00>

    def test_nonsense(self):
        dt = datetime(2012, 1, 1, 11)
>       assert self.formatter._format_token(dt, None) is None

tests/test_formatter.py:166: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <arrow.formatter.DateTimeFormatter object at 0x1025cf670>, dt = datetime.datetime(2012, 1, 1, 11, 0), token = None

    def _format_token(self, dt: datetime, token: Optional[str]) -> Optional[str]:
    
        if token and token.startswith("[") and token.endswith("]"):
            return token[1:-1]
    
        if token == "YYYY":
            return self.locale.year_full(dt.year)
        if token == "YY":
            return self.locale.year_abbreviation(dt.year)
    
        if token == "MMMM":
            return self.locale.month_name(dt.month)
        if token == "MMM":
            return self.locale.month_abbreviation(dt.month)
        if token == "MM":
            return f"{dt.month:02d}"
        if token == "M":
            return f"{dt.month}"
    
        if token == "DDDD":
            return f"{dt.timetuple().tm_yday:03d}"
        if token == "DDD":
            return f"{dt.timetuple().tm_yday}"
        if token == "DD":
            return f"{dt.day:02d}"
        if token == "D":
            return f"{dt.day}"
    
        if token == "Do":
            return self.locale.ordinal_number(dt.day)
    
        if token == "dddd":
            return self.locale.day_name(dt.isoweekday())
        if token == "ddd":
            return self.locale.day_abbreviation(dt.isoweekday())
        if token == "d":
            return f"{dt.isoweekday()}"
    
        if token == "HH":
            return f"{dt.hour:02d}"
        if token == "H":
            return f"{dt.hour}"
        if token == "hh":
            return f"{dt.hour if 0 < dt.hour < 13 else abs(dt.hour - 12):02d}"
        if token == "h":
            return f"{dt.hour if 0 < dt.hour < 13 else abs(dt.hour - 12)}"
    
        if token == "mm":
            return f"{dt.minute:02d}"
        if token == "m":
            return f"{dt.minute}"
    
        if token == "ss":
            return f"{dt.second:02d}"
        if token == "s":
            return f"{dt.second}"
    
        if token == "SSSSSS":
            return f"{dt.microsecond:06d}"
        if token == "SSSSS":
            return f"{dt.microsecond // 10:05d}"
        if token == "SSSS":
            return f"{dt.microsecond // 100:04d}"
        if token == "SSS":
            return f"{dt.microsecond // 1000:03d}"
        if token == "SS":
            return f"{dt.microsecond // 10000:02d}"
        if token == "S":
            return f"{dt.microsecond // 100000}"
    
        if token == "X":
            return f"{dt.timestamp()}"
    
        if token == "x":
            return f"{dt.timestamp() * 1_000_000:.0f}"
    
        if token == "ZZZ":
            return dt.tzname()
    
        if token in ["ZZ", "Z"]:
            separator = ":" if token == "ZZ" else ""
            tz = dateutil_tz.tzutc() if dt.tzinfo is None else dt.tzinfo
            # `dt` must be aware object. Otherwise, this line will raise AttributeError
            # https://github.com/arrow-py/arrow/pull/883#discussion_r529866834
            # datetime awareness: https://docs.python.org/3/library/datetime.html#aware-and-naive-objects
            total_minutes = int(cast(timedelta, tz.utcoffset(dt)).total_seconds() / 60)
    
            sign = "+" if total_minutes >= 0 else "-"
            total_minutes = abs(total_minutes)
            hour, minute = divmod(total_minutes, 60)
    
            return f"{sign}{hour:02d}{separator}{minute:02d}"
    
        if token in ("a", "A"):
            return self.locale.meridian(dt.hour, token)
    
        if token == "W":
            year, week, day = dt.isocalendar()
            return f"{year}-W{week:02d}-{day}"
    
    
>       raise ValueError("FOO BAR")
E       ValueError: FOO BAR

arrow/formatter.py:150: ValueError
```

I don't know when `sub` would ever pass `None` to` _format_token` though... this test case is directly forcing `None` into the `_format_token` function.

---

@isac322 I think a possible solution here could be to simply put a final case in the fall through to just return the empty string. This will take care of all these cases that do not match any of the conditionals in a sane way. It will also allow us to make the return type `str` rather than `Optional[str]`. Thoughts?

isac322: @jadchaar Sounds good! I will add final return statement and remove `Optional`.

isac322: @jadchaar mypy still complains about `Optional`.
```
arrow/formatter.py:125: error: Incompatible return value type (got "Optional[str]", expected "str")
arrow/formatter.py:142: error: Incompatible return value type (got "Optional[str]", expected "str")
```

First one is `return dt.tzname()` case.
I know that `Arrow` always store non naive `datetime` now. but because `DateTimeFormatter` is public, someone could pass naive `datetime` object to `DateTimeFormatter.format()` and in that case `DateTimeFormatter._format_token()` will return `None`.

Second one is `return self.locale.meridian(dt.hour, token)`.


But because those issues are known issue (I only annotate type for those cases), I think we can just leave FIXME comment and cast it as `str` for now. I want to minimize behavior changing of this PR.

jadchaar: Given Isac's latest findings, you think this is good to merge @systemcatch? We might want to consider creating some issues to address these two FIXMEs going forward (might require some significant refactoring) if possible.

systemcatch: Hey @jadchaar I created #918 to track these 2 bugs. We should merge this in now (exciting!).

maroux:
This is missing [py.typed](https://www.python.org/dev/peps/pep-0561/) support which means tools like mypy can't check `arrow` usage. I created #925 for that.

Issue #890, opened 2020-12-01
CI failing on humanize test


jadchaar:
## Issue Description

Current [CRON CI build on master](https://github.com/arrow-py/arrow/runs/1477209342) is failing. The test case `tests/test_arrow.py::TestArrowHumanize::test_day` is failing on the current master between 7:00-7:59pm EST. The tests pass when run at 8pm EST.

```shell
  __________________________ TestArrowHumanize.test_day __________________________
  
  self = <tests.test_arrow.TestArrowHumanize object at 0x7f851fcb8128>
  
      def test_day(self):
      
          later = self.now.shift(days=1)
      
          assert self.now.humanize(later) == "a day ago"
          assert later.humanize(self.now) == "in a day"
      
          # regression test for issue #697
          less_than_48_hours = self.now.shift(
              days=1, hours=23, seconds=59, microseconds=999999
          )
          assert self.now.humanize(less_than_48_hours) == "a day ago"
          assert less_than_48_hours.humanize(self.now) == "in a day"
      
          less_than_48_hours_date = less_than_48_hours._datetime.date()
          with pytest.raises(TypeError):
              # humanize other argument does not take raw datetime.date objects
              self.now.humanize(less_than_48_hours_date)
      
          # convert from date to arrow object
          less_than_48_hours_date = arrow.Arrow.fromdate(less_than_48_hours_date)
  >       assert self.now.humanize(less_than_48_hours_date) == "a day ago"
  E       AssertionError: assert '23 hours ago' == 'a day ago'
  E         - a day ago
  E         + 23 hours ago
  
  tests/test_arrow.py:1851: AssertionError
```

This test case reproduces it locally:

```python
import arrow

now = arrow.Arrow(2020, 12, 1, 0, 55, 46, 798518)
print("now:", now)

less_than_48_hours = now.shift(
    days=1, hours=23, seconds=59, microseconds=999999
)
print("less_than_48_hours:", less_than_48_hours)

less_than_48_hours_date = arrow.Arrow.fromdate(less_than_48_hours._datetime.date())
print("less_than_48_hours_date:", less_than_48_hours_date)

humanized = now.humanize(less_than_48_hours_date)
print("humanized:", humanized)

# Should be "a day ago"
assert humanized == "a day ago"
```

## System Info

- 🖥  **OS name and version**: macOS 11.0 
- 🐍  **Python version**: 3.9.0 
- 🏹  **Arrow version**:  latest `master`

jadchaar:
[...]

PR #906, opened 2020-12-28
Raise exception when None value is passed to arrow.get()

### LINKED ISSUE #900: Should arrow.get(None) work?
As per the title should the following work?

```shell
>>> arrow.get(None)
<Arrow [2020-12-20T20:12:12.631217+00:00]>
```

https://github.com/arrow-py/arrow/blob/master/arrow/factory.py#L169

There could be an issue where someone unintentionally passes `None` in but gets no error. Realistically is this used?

Armaxxx:
## Pull Request Checklist

## Description of Changes

Closes: https://github.com/arrow-py/arrow/issues/900

### DIFF
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -167,6 +167,7 @@ class ArrowFactory:
 
             # (None) -> now, @ utc.
             if arg is None:
+                raise NameError
                 return self.type.utcnow()
 
             # try (int, float) -> from timestamp with tz
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -167,7 +167,7 @@ class ArrowFactory:
 
             # (None) -> raises an exception
             if arg is None:
-                raise TypeError("Cannot parse None.")
+                raise TypeError("Cannot parse argument of type None.")
 
             # try (int, float) -> from timestamp with tz
             elif not isinstance(arg, str) and is_timestamp(arg):
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -168,7 +168,7 @@ class ArrowFactory:
             # (None) -> now, @ utc.
             if arg is None:
                 raise NameError
-                return self.type.utcnow()
+                # return self.type.utcnow()
 
             # try (int, float) -> from timestamp with tz
             elif not isinstance(arg, str) and is_timestamp(arg):
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -165,10 +165,9 @@ class ArrowFactory:
         if arg_count == 1:
             arg = args[0]
 
-            # (None) -> now, @ utc.
+            # (None) -> raises an exception
             if arg is None:
-                raise ValueError
-                # return self.type.utcnow()
+                raise TypeError("Cannot parse None.")
 
             # try (int, float) -> from timestamp with tz
             elif not isinstance(arg, str) and is_timestamp(arg):
--- a/arrow/factory.py
+++ b/arrow/factory.py
@@ -167,7 +167,7 @@ class ArrowFactory:
 
             # (None) -> now, @ utc.
             if arg is None:
-                raise NameError
+                raise ValueError
                 # return self.type.utcnow()
 
             # try (int, float) -> from timestamp with tz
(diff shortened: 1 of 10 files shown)


Armaxxx:
Also I'm not sure how the errors are happening, probably trying to run some None valued get command.

Armaxxx:
I added an assertRaises from unittest

arrow/factory.py
```diff
@@ -167,7 +167,8 @@ def get(self, *args, **kwargs):
 
             # (None) -> now, @ utc.
             if arg is None:
-                return self.type.utcnow()
+                raise NameError
+                # return self.type.utcnow()
```
jadchaar, line 171: A `ValueError` is more appropriate in this situation.

tests/test_factory.py
```diff
@@ -10,7 +11,7 @@
 
 
 @pytest.mark.usefixtures("arrow_factory")
-class TestGet:
+class TestGet(unittest.TestCase):
```
jadchaar, line 14: No need to use `unittest` here. We are using `pytest`, so all you need to do to test this is:

```python
with pytest.raises(ValueError):
    self.factory.get(None)
```

jadchaar (changes requested):

tests/test_factory.py
Armaxxx: How silly of me the example was right below, it has been changed.

arrow/factory.py
Armaxxx: I had my doubts, but you're completely right, it's been changed.

arrow/factory.py
```diff
@@ -167,7 +167,8 @@ def get(self, *args, **kwargs):
 
             # (None) -> now, @ utc.
             if arg is None:
-                return self.type.utcnow()
+                raise ValueError
```
krisfremen, line 170: I'm OK with ValueError, but wouldn't a TypeError be more indicative?

`TypeError("Cannot parse None.")` perhaps?

krisfremen (changes requested):
Few comments.

arrow/factory.py
jadchaar: After reading [this SO post](https://stackoverflow.com/questions/48343387/valueerror-and-typeerror-in-python), I agree with Kris, this should be a `TypeError`. I think Kris's message is sufficient as well.

arrow/factory.py
```diff
@@ -167,7 +167,8 @@ def get(self, *args, **kwargs):
...
+                # return self.type.utcnow()
```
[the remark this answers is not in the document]
Armaxxx: Of course

arrow/factory.py
```diff
@@ -167,7 +167,8 @@ def get(self, *args, **kwargs):
 
             # (None) -> now, @ utc.
```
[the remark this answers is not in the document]
Armaxxx: Yeah makes sense, it now says that it raises an exception

arrow/factory.py
Armaxxx: Great, I used the comment you provided if you don't mind. It's done.

Armaxxx:
Weird, the error is in test-months on humanize class, on the test_arrow.py file, but I didn't fiddle with it.

jadchaar:
@Armaxxx yeah this is due to a recent change to our humanize bounds. @systemcatch ,@krisfremen, and I need to take a closer look (ref https://github.com/arrow-py/arrow/issues/890). Your changes now look good to go though!

krisfremen (approved):
Looking good.

arrow/factory.py
```diff
@@ -165,9 +165,9 @@ def get(self, *args, **kwargs):
         if arg_count == 1:
             arg = args[0]
 
-            # (None) -> now, @ utc.
+            # (None) -> raises an exception
             if arg is None:
-                return self.type.utcnow()
+                raise TypeError("Cannot parse None.")
```
systemcatch, line 170: Maybe here we should say `"Cannot parse argument of type None"`

systemcatch (commented):
Hey @Armaxxx, only one suggestion from me.

jadchaar (approved):

PR #911, opened 2021-01-03
Remove redundant Python2 numeric relics

eumiro:
## Pull Request Checklist

## Description of Changes

Refactor to remove Python2-specific numeric hacks that are no more needed in Python3.

### DIFF
--- a/arrow/formatter.py
+++ b/arrow/formatter.py
@@ -51,16 +51,16 @@ class DateTimeFormatter:
         if token == "MM":
             return f"{dt.month:02d}"
         if token == "M":
-            return str(dt.month)
+            return f"{dt.month}"
 
         if token == "DDDD":
             return f"{dt.timetuple().tm_yday:03d}"
         if token == "DDD":
-            return str(dt.timetuple().tm_yday)
+            return f"{dt.timetuple().tm_yday}"
         if token == "DD":
             return f"{dt.day:02d}"
         if token == "D":
-            return str(dt.day)
+            return f"{dt.day}"
 
         if token == "Do":
             return self.locale.ordinal_number(dt.day)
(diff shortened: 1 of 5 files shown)


arrow/formatter.py
```diff
@@ -70,45 +70,45 @@ def _format_token(self, dt, token):
...
         if token == "HH":
             return f"{dt.hour:02d}"
         if token == "H":
-            return str(dt.hour)
+            return f"{dt.hour}"
         if token == "hh":
-            return "{:02d}".format(dt.hour if 0 < dt.hour < 13 else abs(dt.hour - 12))
+            return f"{dt.hour if 0 < dt.hour < 13 else abs(dt.hour - 12):02d}"
         if token == "h":
             return str(dt.hour if 0 < dt.hour < 13 else abs(dt.hour - 12))
```
jadchaar, line 82: Shouldn't this be changed to an f-string as well?

jadchaar (changes requested):
One small comment.

arrow/formatter.py
eumiro: Right, fixed, repushed, thank you. Your `pre-commit` did not complain, so I ignored that line.

eumiro: That expression `dt.hour if 0 < dt.hour < 13 else abs(dt.hour - 12)` could be written as `(dt.hour - 1) % 12 + 1`, but I am not sure whether you'll like it more than the current solution.

jadchaar: I usually prefer modulo to ternary statements. Any preference @systemcatch?

systemcatch: I tend to prefer ternary but honestly don't mind if you guys want to change it.

jadchaar: Thinking about it, I think I actually agree with Chris here. It is more explicit as it is since the hour range is obvious.

eumiro: This is actually a piece of code that never runs on my computer. A day has only 24±1 hours, not 12 :-)

jadchaar: So you are saying one part of the ternary is dead code?

eumiro: No. I want to say that I personally as a user always use 24-hours format, so this piece of code formatting to 12-hours wouldn't be called on my computer. It was just a fun way to close this discussion, and it has nothing to do with this PR.

I totally accept your preference for the ternary and anyway I am happy I was able to contribute to `arrow`.

jadchaar: Thanks for your contributions we really do appreciate them!

systemcatch (approved):
Ready to go!

systemcatch:
I'll leave merging to @jadchaar in case he has any further comments.

jadchaar (approved):
LGTM

PR #920, opened 2021-02-18
Utilize __init_subclass__ instead of inspect on Locale mapping to improve type checking

isac322:
## Pull Request Checklist

## Description of Changes

Python 3.6 introduced [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html?highlight=__init_subclass__#object.__init_subclass__) that invoked when any subclass is declared. It can replace behavior of `arrow.locales._map_locales()`.

### Additional feature

`__init_subclass__` invoked on **every subclassing** -> Subclass of `Locale` outside of `arrow` package also registered on `arrow.locales._locale_map`.

#### Example

```
>>> from arrow.locales import Locale
>>> class Locale1(Locale):
...   names = ['asdf']
... 
>>> from arrow import locales
>>> locales.get_locale('asdf')
<__main__.Locale1 object at 0x100f19b50>
```


arrow/locales.py
```diff
@@ -121,6 +123,13 @@ class Locale:
 
     _month_name_to_ordinal: Optional[Dict[str, int]]
 
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        for locale_name in cls.names:
+            if locale_name in _locale_map:
+                raise LookupError(f"Duplicated locale name: {locale_name}")
```
isac322, line 129: Should I handle duplicated key case like this?

krisfremen: Hmm, that would be a nice way of doing it. But from my POV, shouldn't we test for duplicates as part of a merge request, so we don't end up with duplicates instead of run time?

krisfremen (commented):
One comment from me, otherwise, I don't see anything wrong from a first glance.

Will take a better look later today.

arrow/locales.py
```diff
@@ -121,6 +123,13 @@ class Locale:
 
     _month_name_to_ordinal: Optional[Dict[str, int]]
 
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        for locale_name in cls.names:
+            if locale_name in _locale_map:
```
jadchaar, line 128: Should we maintain the `locale_name.lower()` like in the old code?

jadchaar (approved):
LGTM but had a small comment

PR #925, opened 2021-02-26
Add py.typed export per PEP 561

maroux:
## Pull Request Checklist

## Description of Changes

Add `py.typed` file and export type annotations per [PEP 561](https://www.python.org/dev/peps/pep-0561/) so that type testing tools like mypy can actually verify types when using arrow.

No new test added for this change and I don't think there's any documentation necessary but let me know if something needs to be added.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ setup(
     author_email="crsmithdev@gmail.com",
     license="Apache 2.0",
     packages=["arrow"],
+    package_data={"arrow": ["py.typed"]},
     zip_safe=False,
     python_requires=">=3.6",
     install_requires=[


setup.py
```diff
@@ -19,6 +19,7 @@
     author_email="crsmithdev@gmail.com",
     license="Apache 2.0",
     packages=["arrow"],
+    package_data={"arrow": ["py.typed"]},
```
jadchaar, line 22: Might be better to specify this in `MANIFEST.in`: https://github.com/arrow-py/arrow/blob/master/MANIFEST.in

jadchaar: Actually looking at PEP 561, this seems like the way to specify it. Disregard.

jadchaar (approved):
LGTM, but I had a quick question @maroux. If PEP 561 is Python 3.7+ compatible, will mypy not detect the typings in Python 3.6?

maroux:
@jadchaar possible? At least for mypy, it looks like it supports `py.typed` even for py3.6:
```
○ → cat a.py
import arrow


if __name__ == "__main__":
    print(arrow.get(object()))


○ → python --version
Python 3.6.11


○ → mypy a.py
a.py:5: error: No overload variant of "get" matches argument type "object"
a.py:5: note: Possible overload variant:
a.py:5: note:     def get(Union[Arrow, datetime, date, struct_time, tzinfo, int, float, str, Tuple[int, int, int]], *, locale: str = ..., tzinfo: Union[tzinfo, str, None] = ..., normalize_whitespace: bool = ...) -> Arrow
a.py:5: note:     <3 more non-matching overloads not shown>
Found 1 error in 1 file (checked 1 source file)

```

I think for PEPs that don't add features to python core, the `python-version` isn't a requirement, but I could be wrong.

jadchaar:
@isac322 you might be more familiar with PEP 561 for type annotations. Think this is good to merge?

isac322:
@jadchaar Yep I forgot to add this 😅

jadchaar:
> @jadchaar Yep I forgot to add this 😅

All good, just wanted to make sure!

Issue #952, opened 2021-04-07
Standardize separator and casing used in locale names


jadchaar:
The separator we use for locale names seems to be inconsistent. Some locales like Swahili use the underscore for their locale names: https://github.com/arrow-py/arrow/blob/d515e7463700b7da4c89caf81bcb736395abf898/arrow/locales.py#L4308-L4314. Others like Croatian use the hyphen: https://github.com/arrow-py/arrow/blob/d515e7463700b7da4c89caf81bcb736395abf898/arrow/locales.py#L4393-L4395.

As for casing, Hebrew seems to capitalize `IL`, but others do all lowercase: https://github.com/arrow-py/arrow/blob/d515e7463700b7da4c89caf81bcb736395abf898/arrow/locales.py#L3225-L3227.

These should be standardized and validated in a test case.

jadchaar:
We should be using dashes internally in `locales.py` since that seems to align with the ISO standard: https://en.wikipedia.org/wiki/Language_code. That said, we need to replace underscores on locale inputs to ensure backward compatibility.

PR #954, opened 2021-04-08
Standardize locales and improve locale validation

jadchaar:
## Pull Request Checklist

## Description of Changes
Closes: https://github.com/arrow-py/arrow/issues/952

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,7 +105,7 @@ Example Usage
     >>> local.humanize()
     'an hour ago'
 
-    >>> local.humanize(locale='ko_kr')
+    >>> local.humanize(locale='ko-kr')
     '1시간 전'
 
 .. end-inclusion-marker-do-not-remove
--- a/README.rst
+++ b/README.rst
@@ -106,7 +106,7 @@ Example Usage
     'an hour ago'
 
     >>> local.humanize(locale='ko-kr')
-    '1시간 전'
+    '한시간 전'
 
 .. end-inclusion-marker-do-not-remove
 
--- a/arrow/api.py
+++ b/arrow/api.py
@@ -22,7 +22,7 @@ _factory = ArrowFactory()
 @overload
 def get(
     *,
-    locale: str = "en_us",
+    locale: str = "en-us",
     tzinfo: Optional[TZ_EXPR] = None,
     normalize_whitespace: bool = False,
 ) -> Arrow:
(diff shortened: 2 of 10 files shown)


jadchaar:
@systemcatch @krisfremen do ya'll think I should change the variable that is referenced when we throw errors about locales? E.g. duplicate locale registered in a subclass or if a locale does not exist in a call to `get_locale`? Currently, we use the original locale they pass in like `en_US`, but if we are transforming it into `en-us` for our internal lookup, should we display this in the error message instead?

krisfremen (commented):
Since we are doing a bit of sanitizing anyway, I think what we have in the locales should be fine to return in the errors, it'll help to adopt the new codes also.

systemcatch:
I agree with Kris, let's start returning the new codes in errors.

jadchaar:
Fixed @krisfremen @systemcatch!

arrow/api.py
```diff
@@ -22,7 +22,7 @@
 @overload
 def get(
     *,
-    locale: str = "en_us",
+    locale: str = "en-us",
```
krisfremen, line 25: Should we maybe put this in the constants?

DEFAULT_LOCALE?

krisfremen (commented):
One question from first glance, otherwise looking good.

README.rst
```diff
@@ -105,7 +105,7 @@ Example Usage
     >>> local.humanize()
     'an hour ago'
 
-    >>> local.humanize(locale='ko_kr')
+    >>> local.humanize(locale='ko-kr')
     '1시간 전'
```
krisfremen, line 109: I think this is not the output currently, since the last changes to the Korean locale by comfuture.

krisfremen (commented):
One more thought off the top of my head, but I'll have to double check this.

README.rst
krisfremen: Indeed.

```
>>> import arrow
>>> kr = arrow.utcnow().shift(hours=-1)
>>> kr.humanize(locale='ko_kr')
'한시간 전'
```

jadchaar: Fixed

arrow/api.py
jadchaar: Agreed. Done!

krisfremen (approved):
Looks good to me. 👍

jadchaar:
Look good to you @systemcatch?

systemcatch:
Yeah Jad looks good to merge.

PR #895, opened 2020-12-11
Dehumanize locale attempt

egannon:
## Pull Request Checklist

Thank you for taking the time to improve Arrow! Before submitting your pull request, please check all *appropriate* boxes:

- [X] 🧪  Added **tests** for changed code.
- [X] 🛠️  All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [ ] 🧹  All linting checks **pass** when run locally (run `tox -e lint` or `make lint` to find out!).
- [ ] 📚  Updated **documentation** for changed code.
- [ ] ⏩  Code is **up-to-date** with the `master` branch.

If you have *any* questions about your code changes or any of the points above, please submit your questions along with the pull request and we will try our best to help!

## Description of Changes

With limited time to work on this dehumanize feature, we only tackled aspects of integrating dehumanize with locales. We don’t intend this PR to be merged, but simply to serve as a basis for a contributor who may work on this task in the future. With our current implementation, we worked off of the code that was functioning for the English locale (PR request #772). Our main contribution was to attempt to translate any other locale that is received as an input into English. For this reason, we added an input parameter locale. Additionally, our translation code required us to utilize new dictionaries from the locales.py. For the languages we tested on we added a revered_timeframes dict and a numbers dict (with only 1-10 for testing purposes). Our current implementation works for the tests we created with languages: French, Spanish, Hindi, and German. 

There are many limitations to the current implementation:

Dehumanize requires an existing arrow object.
“Time in locale_data.past” doesn’t work with languages whose past  (or future) variables in locale.py contain {0} in the middle of the string.
Languages that don’t separate by space. 
Creating reversed_timeframe dict and numbers dict for each locale would be tedious. 

Although we see many potential obstacles with a continuation of our implementation, we hope this PR can help induce further ideas or contributions to this de-humanize issue. Thank you!

Closes: #360

PR #956, opened 2021-04-08
Implement dehumanize method

anishnya:
## Pull Request Checklist

- [ x] 🧪  Added **tests** for changed code.
- [ x] 🛠️  All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [ x] 🧹  All linting checks **pass** when run locally (run `tox -e lint` or `make lint` to find out!).
- [ x] 📚  Updated **documentation** for changed code.
- [ x] ⏩  Code is **up-to-date** with the `master` branch.

## Description of Changes
This PR isn't expected to be merged, however, we would like to start a dialogue on how to proceed with some of the challenges ahead in implementing a reverse_humanize/dehumanize method as proposed in #360. We would first like to thank #772 and #895 for giving us a good base to work off of. 

This attempt so far works with 37 of the 54 current locales in Arrow, a big step forward from previous attempts. There are a couple issues we would like to address and discuss. 

**1) Input Validation**

Currently, this solution has little input validation. For example, a user can "in 3 months and 3 lightseconds" and dehumanize will return an object that is 3 months ahead. After some experimenting, creating a generalized method for input validation across locales seems unfeasible. 

For example, one method we attempted was to simply delete any match from the input string and then at the very end of search, raise an error if the string was not empty (after removing the locale's and word and removing any whitespace). This creates some issues in languages such as Azerbaijani, the plural for a unit is the same as a single value for a unit. For example, the phrase for "a second" in Azerbaijani is "saniyə" whereas the phrase for "5 seconds" is "5 saniyə". This type of situation has only been encountered in 3 languages so far. There is an easy fix here, but we worry there may be another locale introduced that breaks those rules entirely. 

Some dialogue is needed here on how much input validation is or isn't needed. Potentially we can label dehumanize as fuzzy_dehumanize in the interim and revisit the question of input validation later. 

**2) Remaining Locales and the Need for Standardization Across Locales** 

In order to support the remaining locales not supported by this feature, there needs to be some element of standardization across locales. This pertains to what we will be referring to as special time frame values (a special word conveying two days ago). 

Some locales incorporate these special time frame values very differently. Some like the Slavic languages used the regular timeframe object and have the key as some list. 

Bulgarian:
`"minutes": ["{0} минута", "{0} минути", "{0} минути"]`

Some have their timeframe key as a dictionary. 

Arabic:
`"seconds": {"double": "ثانيتين", "ten": "{0} ثوان", "higher": "{0} ثانية"},`

Some have some nested combination of both. 

Czech:
`"seconds": {"past": "{0} sekundami", "future": ["{0} sekundy", "{0} sekund"]}`

Some introduce their own separate objects.

Korean:
`special_dayframes = {
        -3: "그끄제",
        -2: "그제",
        -1: "어제",
        1: "내일",
        2: "모레",
        3: "글피",
        4: "그글피",
    }`

A temporary solution here would be to create a special time frame dictionary for each of these locales, allowing dehumanize to function on these languages and then later down the line figuring out a way to standardize across locales.
A more permanent solution maybe to standardized like the following. 

Temporary Solution:
`special_timeframes = {"string for two months ago": ["months", 2, True], "{0} special plural unit": ["seconds", 1, False]}`

The first element in the list represents the unit that string modifies, the second the value modified and the third is a bool if we need to obtain the value from the string or not. 

Permanent Solution:
`"seconds": {1: "special value one", 5: "special value two", "general": "general value one {0}", "special case one": "special case string {0}"}`

We could check if a key is an int and know the value associated with that phrase for the particular unit. If the key is some string, we know there the value that can be associated with that phrase is arbitrary and we need to obtain the value.

I believe these issues aren't too complex to solve, but I think we need some discussion to choose from multiple potential soltuons.


systemcatch:
Hey @anishnya from a quick skim over (more detailed review will follow) I really like this PR and think merging it is quite possible. If it only works for 37/54 locales at the moment we can encode a locale check and work on improving the method. With dehumanize and all the PR attempts before I think we've let perfect be the enemy of good sometime.

arrow/arrow.py
```diff
@@ -1282,6 +1283,102 @@ def gather_timeframes(_delta: float, _frame: TimeFrameLiteral) -> float:
...
+
+        return current_time.shift(
+            seconds=sign_val * time_object_info["seconds"],
+            minutes=sign_val * time_object_info["minutes"],
+            hours=sign_val * time_object_info["hours"],
+            days=sign_val * time_object_info["days"],
+            weeks=sign_val * time_object_info["weeks"],
+            months=sign_val * time_object_info["months"],
+            years=sign_val * time_object_info["years"],
+        )
```
systemcatch, line 1380:
```suggestion

        time_changes = {k: sign_val*v for k, v in time_object_info.items()}

        return current_time.shift(**time_changes)
```

Or something along those lines, saves having to do all the multiplications in the return.

arrow/arrow.py
```diff
@@ -1282,6 +1283,102 @@ def gather_timeframes(_delta: float, _frame: TimeFrameLiteral) -> float:
...
+        past_pattern_match = past_pattern.findall(timestring)
+
+        if past_pattern_match:
+            sign_val = -1
+
+        elif future_pattern_match:
+            sign_val = 1
+
+        else:
+            raise ValueError(
```
systemcatch, line 1331: We need lots of tests where garbage/bad info is passed to `dehumanize`.

systemcatch (changes requested):
Hey @anishnya I've done a first pass with some comments. I think you should create a list of supported locales in `constants.py` and perform a check with the provided locale, raising an error if it's not supported.

arrow/arrow.py
```diff
@@ -1282,6 +1283,102 @@ def gather_timeframes(_delta: float, _frame: TimeFrameLiteral) -> float:
...
+        # Create an object containing the relative time info
+        time_object_info = {
+            "seconds": 0,
+            "minutes": 0,
+            "hours": 0,
+            "days": 0,
+            "weeks": 0,
+            "months": 0,
+            "years": 0,
+        }
```
systemcatch, line 1303:
```suggestion
        time_object_info = dict.fromkeys(["seconds", "minutes", "hours", "days", "weeks", "months", "years"], 0)
```
Purely personal preference, feel free to ignore.

anishnya:
Will get on these issues asap @systemcatch. How much input validation is needed? Just input that is completely bad, like for example a random string? Or more nuanced input validation like rejecting "2 seconds 10 seconds ago"? As mentioned in the original PR post, the latter is proving to be quite difficult to generalize across locales. The current solution will reject any string that does not contain some kind of time reference like "ago" or "in".

systemcatch:
Let's start with simple validation and work from there.

anishnya:
I've implemented all the suggestions and added additional input validation test cases. Although the input validation isn't perfect, it does reject complete nonsense. Apologies in advance for the weird merge issues in this PR. I'm not too sure why that ended up happening.

systemcatch:
> I've implemented all the suggestions and added additional input validation test cases. Although the input validation isn't perfect, it does reject complete nonsense. Apologies in advance for the weird merge issues in this PR. I'm not too sure why that ended up happening.

Not to worry we'll squash all the commits to make it look nice and clean.

arrow/arrow.py
```diff
@@ -1282,6 +1299,132 @@ def gather_timeframes(_delta: float, _frame: TimeFrameLiteral) -> float:
...
+                change_value = 1
+            else:
+                change_value = int(num_match.group())
+
+            # No time to update if now is the unit
+            if unit == "now":
+                unit_visited[unit] = True
+                continue
+
+            # Add change value to the correct unit (incorporates the plualirty that exists within timeframe i.e second v.s seconds)
```
systemcatch, line 1385: plurality

arrow/arrow.py
```diff
@@ -1282,6 +1299,132 @@ def gather_timeframes(_delta: float, _frame: TimeFrameLiteral) -> float:
...
+            time_unit_to_change = str(unit)
+            time_unit_to_change += "s" if (str(time_unit_to_change)[-1] != "s") else ""
+            time_object_info[time_unit_to_change] = change_value
+            unit_visited[time_unit_to_change] = True
+
+        # Assert error if string does not modify any units
+        if not any([True for k, v in unit_visited.items() if v]):
+            raise ValueError(
+                """Input string not valid. Note: Some locales do not support the week granulairty in Arrow.
+                 If you are attempting to use the week granulairty on an unsupported locale, this could be the cause of this error."""
```
systemcatch, line 1395: granularity

systemcatch (changes requested):
Looking good @anishnya, I've got a couple more comments which is mostly small stuff.

We'll need dehumanize to be mentioned in the documentation, how about we put one or two examples after the [humanize](https://arrow.readthedocs.io/en/latest/#humanize) section?

anishnya:
> Looking good @anishnya, I've got a couple more comments which is mostly small stuff.
> 
> We'll need dehumanize to be mentioned in the documentation, how about we put one or two examples after the [humanize](https://arrow.readthedocs.io/en/latest/#humanize) section?

I've pushed the suggest changes and added the examples to the docs.

docs/index.rst
```diff
@@ -252,6 +252,41 @@ Support for a growing number of locales (see ``locales.py`` for supported langua
     >>> future.humanize(a, locale='ru')
     'через 2 час(а,ов)'
 
+Dehumanize
+~~~~~~~~
+
+Dehumanize a string representing a past time:
```
systemcatch, line 258: How about `"Take a human readable string and use it to shift into a past time."`

systemcatch (changes requested):

docs/index.rst
```diff
@@ -252,6 +252,41 @@ Support for a growing number of locales (see ``locales.py`` for supported langua
     >>> future.humanize(a, locale='ru')
     'через 2 час(а,ов)'
 
+Dehumanize
+~~~~~~~~
```
systemcatch, line 256: The underline length needs to match the title (causes the lint action to fail).

systemcatch (changes requested):
One last tweak and we should be good to merge @anishnya

docs/index.rst
anishnya: Linting issue fixed, as well as documentation updated.

systemcatch (approved):
Nicely done @anishnya, thanks for your effort on this.

Issue #977, opened 2021-05-13
Add support for Excel date format


PyGeek03:
## Feature Request
In Excel files, date is represented as an integer that counts how many days have passed from [either January 1st, 1900 (on Windows) or January 1st, 1904 (on Mac)](https://docs.microsoft.com/en-US/office/troubleshoot/excel/1900-and-1904-date-system). 

The function ([xlrd.xldate.xldate_as_datetime](https://xlrd.readthedocs.io/en/latest/api.html#xlrd.xldate.xldate_as_datetime) from the xlrd library is usually used for parsing this date format in Python. Considering Arrow's goal of providing a common API for date parsing, I believe this would be a useful feature for users who have to wrangle multiple date formats in the same column of an Excel file (because of non-standard data entry). 

The only issue is, I'm not sure if this format should be supported as a token to use with arrow.get() and arrow.format() functions (and what that token should be), or if it should be its own function. I'm happy to work on this once we've decided on whether to implement this and how to solve this issue.

anishnya:
Hi @PyGeek03. Do you mind giving us some examples of inputs and expected outputs?

PyGeek03:
> Hi @PyGeek03. Do you mind giving us some examples of inputs and expected outputs?

These examples are from the [Microsoft documentation page](https://docs.microsoft.com/en-US/office/troubleshoot/excel/1900-and-1904-date-system):

- For the 1900 date system (starts counting from January 1, 1900):
Input: 0. Output: January 1, 1900
Input: 34519. Output: July 4, 1994
Input: 35981. Output: July 5, 1998

- For the 1904 date system (starts counting from January 1, 1904):
Input: 0. Output: January 1, 1904
Input: 34519. Output: July 5, 1998
Input: 35981. Output: July 6, 2002

An interesting issue is that [for compatibility with Lotus 1-2-3, Excel considers 1900 to be a leap year, which is wrong](https://docs.microsoft.com/en-US/office/troubleshoot/excel/wrongly-assumes-1900-is-leap-year). So the difference between the 2 date systems is not 4 years, but 4 years 1 day.

anishnya:
Hi @PyGeek03, I was looking a little more into the documentation. Is the input ever a fraction, or float? I just want to make sure there aren't any weird edge cases I'm missing before I propose a solution.  Some of our dependencies don't play nicely with floats.

PyGeek03:
[...]

PR #979, opened 2021-05-26
Creation from Excel Serial Dates

anishnya:
## Pull Request Checklist

- [ x] 🧪  Added **tests** for changed code.
- [ x] 🛠️  All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [ x] 🧹  All linting checks **pass** when run locally (run `tox -e lint` or `make lint` to find out!).
- [ x] ⏩  Code is **up-to-date** with the `master` branch.

## Description of Changes

Closes: #977. New support for creation of Arrow objects from Excel Serial Dates. There may be a linting issue, with the doc string that needs to be addressed.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -32,6 +32,7 @@ from typing import (
 from dateutil import tz as dateutil_tz
 from dateutil.relativedelta import relativedelta
 
+import arrow
 from arrow import formatter, locales, parser, util
 from arrow.constants import DEFAULT_LOCALE, DEHUMANIZE_LOCALES
 from arrow.locales import TimeFrameLiteral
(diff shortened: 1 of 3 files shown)


arrow/arrow.py
```diff
@@ -1787,6 +1788,42 @@ def _get_tzinfo(tz_expr: Optional[TZ_EXPR]) -> dt_tzinfo:
             except parser.ParserError:
                 raise ValueError(f"{tz_expr!r} not recognized as a timezone.")
 
+    @classmethod
+    def excel_date(
+        cls, delta: Union[int, float], default_windows_date: bool = True
```
jadchaar, line 1793: Should we change the kwarg to instead use the Microsoft terminology of 1900 and 1904 datesystem: https://docs.microsoft.com/en-us/office/troubleshoot/excel/1900-and-1904-date-system? Like make the kwarg `1904_date_system=True`? Or do you think using the OS names is clearer?

arrow/arrow.py
```diff
@@ -1787,6 +1788,42 @@ def _get_tzinfo(tz_expr: Optional[TZ_EXPR]) -> dt_tzinfo:
...
+        """
+
+        if default_windows_date:
+            # Need to have this clause as Excel incorrectly considers 1900 a leap year
+            if delta < 60:
+                start_date = arrow.get("1899-12-31")
+            else:
+                start_date = arrow.get("1899-12-30")
+        else:
+            start_date = arrow.get("1904-01-01")
```
jadchaar, line 1821: Instead of using `arrow.get()` you can call `cls.get()`.

jadchaar (changes requested):
Couple of comments

arrow/arrow.py
anishnya: I think it makes more sense to use the 1900/1904 convention. I looked in Excel and that's what they end up of using.

krisfremen (commented):
Would this PR be incorporating the decimal portion of the excel serial?

anishnya:
> Would this PR be incorporating the decimal portion of the excel serial?

Yes it would. I can add test cases to make it more clear.

jadchaar:
Rather than creating a new function, I think it may be more consistent with our existing `get()` interface to add this as a keyword argument like `arrow.get(..., excel: bool, default_windows_date: bool)`. If the excel kwarg is present, we parse the timestamp as an excel timestamp (and account for the `default_windows_date` kwarg as well), else just parse the input time string as normal.

systemcatch:
@jadchaar do you think that `arrow.get()` might have too many kwargs if we add this?

Issue #961, opened 2021-04-25
Improve locale support for dehumanize


systemcatch:
Now that #956 is merged let's focus on getting more locales to work with dehumanize.

- list of unsupported locales to follow....

anishnya:
I think in general there needs to be some clean up of legacy locales. From my experience, there is too much variation between their implementations. Granted, this task may end up being extremely difficult given that some of these locales haven't been touched in years. 

In my opinion, the best plan to get dehumanize to 100% locale support is to first standardize locale implementations (we may want to open this up as a separate issue).  Some ways to do this were suggested in #956. I can propose more detailed solutions in the coming weeks and we can discuss.

PR #1017, opened 2021-08-07
Refactor of Icelandic Locale For Dehumanize Support

anishnya:
## Pull Request Checklist

- [ ] 🛠️  All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [ ] ⏩  Code is **up-to-date** with the `master` branch.

## Description of Changes

Refactored Icelandic Locale for Future Dehumanize Support (#961 and #970).

### DIFF
--- a/arrow/locales.py
+++ b/arrow/locales.py
@@ -2759,8 +2759,13 @@ class IcelandicLocale(Locale):
         if isinstance(form, Mapping):
             if delta < 0:
                 form = form["past"]
-            else:
+            elif delta > 0:
                 form = form["future"]
+            else:
+                raise ValueError(
+                    "Icelandic Locale does not support units with a delta of zero. "
+                    "Please consider making a contribution to fix this issue."
+                )
                 # FIXME: handle when delta is 0
 
         return form.format(abs(delta))
--- a/arrow/locales.py
+++ b/arrow/locales.py
@@ -2759,11 +2759,11 @@ class IcelandicLocale(Locale):
         if isinstance(form, Mapping):
             if delta < 0:
                 form = form["past"]
-            elif delta > 0:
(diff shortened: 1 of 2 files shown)


arrow/locales.py
```diff
@@ -2755,11 +2755,13 @@ class MoroccoArabicLocale(ArabicLocale):
...
-        elif delta > 0:
-            form = form[1]
-            # FIXME: handle when delta is 0
+
+        if isinstance(form, Mapping):
+            if delta < 0:
+                form = form["past"]
+            elif delta >= 0:
+                form = form["future"]
+                # FIXME: handle when delta is 0
```
systemcatch, line 2764: I guess this is now fixed when `delta=0`?

anishnya: So I think there are two options here. We default to the future value when the case is that the delta for a particular unit is zero, or we implicitly error out like we currently do. In an ideal scenario, I would add a raise value error here, but I fear that would probably break the use of the locale, given how common the use of a delta of zero is.

systemcatch: I would tend to think error out is better as it is probably less surprising.

systemcatch:
```shell
(arrow) chris@Z490:~/arrow$ python
Python 3.9.1 (default, Feb 17 2021, 16:56:42) 
[GCC 10.2.0] on linux
>>> import arrow
>>> start=arrow.utcnow()
>>> start.humanize(other=start, locale="is")
'rétt í þessu'
>>> start.humanize(other=start, granularity="hour",locale="is")
'eftir 0 tíma'
```

systemcatch (approved):

krisfremen (approved):
LGTM

PR #1018, opened 2021-08-08
Add an "all" granularity to humanize

### LINKED ISSUE #1014: Add an "all" granularity to humanize
## Feature Request

Once #997 is implemented, using full precision would become more viable. However, it's currently cumbersome to specify that; it requires a list of all units. Sure, a wrapper can be made to abstract that away, but it'd be convenient to just pass the string `"all"`.

MarkKoz:
## Pull Request Checklist

## Description of Changes

Add an "all" granularity to the humanize function. When specified, it internally sets all possible units for the granularity (years, months, weeks, days, hours, minutes, and seconds). This granularity is mutually exclusive with the other granularities; it's not valid to specify it in a list of granularities.

The motivation is that it's more convenient to specify a single string than a list of 7 strings. It will make more sense to use full precision once #997 is merged. Speaking of which, some newly added tests will have to be fixed once #997 is merged and the `describe_multi` bug is fixed.

Resolve #1014

### DIFF
--- a/tests/test_arrow.py
+++ b/tests/test_arrow.py
@@ -2032,6 +2032,49 @@ class TestArrowHumanize:
             == "a minute and 2 seconds ago"
         )
 
+    def test_all_granularity(self):
+        assert (
+            self.now.humanize(granularity="all")
+            == "in 0 years 0 months 0 weeks 0 days 0 hours 0 minutes and 0 seconds"
+        )  # TODO: this should be "ago"; change this when #997 is merged
+
+        later105 = self.now.shift(seconds=10 ** 5)
+        assert (
+            self.now.humanize(later105, granularity="all")
+            == "0 years 0 months 0 weeks a day 3 hours 46 minutes and 40 seconds ago"
+        )
+        assert (
+            later105.humanize(self.now, granularity="all")
+            == "in 0 years 0 months 0 weeks a day 3 hours 46 minutes and 40 seconds"
+        )
+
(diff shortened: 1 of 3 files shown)


tests/test_arrow.py
```diff
@@ -2032,6 +2032,49 @@ def test_multiple_granularity(self):
             == "a minute and 2 seconds ago"
         )
 
+    def test_all_granularity(self):
+        assert (
+            self.now.humanize(granularity="all")
+            == "in 0 years 0 months 0 weeks 0 days 0 hours 0 minutes and 0 seconds"
```
systemcatch, line 2038: Could this cause an issue by not having the `other` parameter set like the rest of the tests?

MarkKoz: I'm not sure. It might technically be time-sensitive, but in practice it probably won't take >= 1 second. For what it's worth, this other test does essentially the same thing:

https://github.com/arrow-py/arrow/blob/89f77b15a27d57db6ff0b8bf013decf3d970574e/tests/test_arrow.py#L1965-L1967

MarkKoz: I fixed this anyway in 804eece.

MarkKoz:
The lint check failure seems to be a problem with CI rather than with my changes.

systemcatch:
Hmmm lint passes now, looks like it was some kind of `setuptools` install error that was causing issues.

jadchaar:
Hi @MarkKoz mind resolving the conflicts in this PR?

Issue #856, opened 2020-09-28
Add flake8 checks for annotations


jadchaar:
## Feature Request

As part of the Python 2.7/3.5 deprecation PR https://github.com/arrow-py/arrow/pull/800, we are adding type annotations. It would be nice to add a flake8 plugin that helps with annotations (e.g. reminds us to place them, static type checking, etc).

jadchaar:
This is the flake8 package we want: https://pypi.org/project/flake8-annotations/.

PR #1061, opened 2021-10-28
add flake8-annotations package for type linting

nthall:
- add pkg in requirements-dev.txt
    - require pkg in pre-commit config
    - add some easily resolved annotations in tests
    - ignore some type warnings in tox.ini

## Pull Request Checklist

## Description of Changes

Closes: #856 
This change adds the `flake8-annotations` plugin to enable linting of type annotations via `pre-commit`/`tox`. In order to enable that without breaking the build, I've added some annotation warnings to the flake8 ignore lists. I thought it would best be left to the maintainers to decide whether it would be more appropriate to address those warnings in separate commits/PRs. 

The ignored warnings, and some brief discussion:

- Globally ignored:
  - `ANN101 Missing type annotation for self in method` / `ANN102 Missing type annotation for cls in classmethod`
    - [PEP 484](https://www.python.org/dev/peps/pep-0484/#id34) says, "In most cases the first argument of class and instance methods does not need to be annotated, and it is assumed to have the type of the containing class for instance methods, and a type object type corresponding to the containing class object for class methods." `flake8-annotations` warns on this case anyway; given the large number of changes required to "fix" this warning I thought it better here to ignore them and have a smaller commit that still advanced the state of the project. Of course if they're desired it would be cleaner to address that issue separately from whether to lint for annotations at all.
- Ignored only in the `tests` dir:
  - `ANN201 Missing return type annotation for public function`
    - Adding such annotations to all tests would again make the commit pretty large.
  - `ANN001 Missing type annotation for function argument`
    - This warning appears throughout the tests, including in the `pytest` fixtures in `conftest.py`; again, seems like a change worth discussing separately from existing annotations.
  - HOWEVER, I did add a very few annotations in the test suite nonetheless, as they seemed to me not to be very intrusive.


nthall:
@jadchaar thanks - sorry for that. I've pushed a fix. Frankly one reason I wanted to tackle this was to improve my familiarity with type annotations, as most of the code I get to work on right now is still on 2.7 with no plans for upgrade or mypy integration.

nthall:
Ah, this is a tougher one. Down in the dependency chain for `flake8-annotations` is `typed-ast`, which has [no plans to support pypy](https://github.com/python/typed_ast/issues/111).  I'm not sure what, if any, resolution would make sense here.

jadchaar:
> Ah, this is a tougher one. Down in the dependency chain for `flake8-annotations` is `typed-ast`, which has [no plans to support pypy](https://github.com/python/typed_ast/issues/111). I'm not sure what, if any, resolution would make sense here.

Since we are only running `flake8` via `pre-commit` we should be able to remove `flake8-annotations` from the `requirements-dev.txt`, which should allow pypy to build successfully again.

nthall:
OK. I've updated `requirements-dev.txt`.

Happy Halloween!

nthall:
I think this is done now? The only thing arguably left would be to rebase it into a cleaner commit - if you want me to do that, I will.

anishnya:
> I think this is done now? The only thing arguably left would be to rebase it into a cleaner commit - if you want me to do that, I will.

No need for the cleaner commit. We can do a simple squash on it. The PR looks good to me, but I'm just waiting on approval from @jadchaar or @krisfremen before we pull it in.

jadchaar (approved):

jadchaar:
Looks great thanks for the awesome work @nthall!

PR #1102, opened 2022-04-05
Require types-python-dateutil

tbekolay:
## Pull Request Checklist

`tox` passes, but when I run `tox -e lint` I get a number of `ANN401 Dynamically typed expressions` errors.

I'm happy to add a test for the change but it's sort of unclear how to do this since it isn't a runtime change, it's a change that should only be picked up when running `mypy`. A short snippet that will raise an error without installing `types-python-dateutil` is:

```
(arrow.now() - arrow.now()).total_seconds()

```

## Description of Changes

Not having `types-python-dateutil` installed results in confusing errors when running `mypy` on the subtraction of two `Arrow` instances. See https://github.com/python/mypy/issues/11613 for a description of the issue, as well as some debugging from a few people.


krisfremen (commented):
@anishnya @jadchaar @systemcatch 

would setup.py be the best place to put this?
or maybe the requirements files would be a better place?

tbekolay:
> or maybe the requirements files would be a better place?

The requirements file does not get used when installing `arrow` through normal means like `pip install arrow`, so putting it there would not fix the issue I reference in the PR description.

jadchaar:
Hi all thanks for the contribution here. Given that this PR now has conflicts, I have made the change in https://github.com/arrow-py/arrow/commit/3a6cd95389dfe1c93684aa083a6e3db408269fdb. Thanks a ton for raising this @tbekolay

tbekolay:
Awesome thanks @jadchaar!

PR #1142, opened 2022-11-14
Force exact match on .get()

### LINKED ISSUE #1084: force exact match on .get()
## Issue Description

when executing the following statement:

```
import arrow
arrow.get('95-03-10 23:01:44','YY-M-D HH:mm:ss')
<Arrow [1995-03-10T23:01:44+00:00]>
```

the library returns me the date, but I gave it a single M, so it shouldn't have done the .get() but it did. one would expect it to fail, as the correct format should be 'YY-MM-D HH:mm:ss' for that same example.

I would like to know if this is a bug or is there a parameter to force the match to be exact?

## System Info

- 🖥  ubuntu 18.04 32 bits
- 🐍  Python 3.8.5
- 🏹 1.2.1

13MK3:
## Pull Request Checklist

## Description of Changes

These changes were made in collaboration with @khanm3

- Modified the _ONE_OR_TWO_DIGIT_RE regex to not accept numbers with leading zeros. This is done in order to conform to the expected behavior and throw an error when, for example, you try to match '01' with 'M'. The issue #1084 only mentions the 'M' token, but realized that the 'D', 'H', 'm', 'h', and 's' have similar behavior which we have also corrected.
- Modified an existing test case since this test was using the undesired behavior which this pull request is intended to fix
- Added a test case which checks the expected behavior for all 6 tokens.

Closes: #1084

Questions:
- We also noticed that the 'DDD' token seems to have similar behavior as well. If you try to match '001' to 'DDD', should that also be expected to fail? We left that one untouched since it uses a different regex and we weren't positive.
- Unrelated to this pull request, but we noticed that tests/test_arrow.py::TestArrowHumanize::test_months is marked as expected to fail, but before we made any changes, it was already passing. Maybe this should be updated?

### DIFF
--- a/arrow/parser.py
+++ b/arrow/parser.py
@@ -107,7 +107,7 @@ class DateTimeParser:
     )
     _ESCAPE_RE: ClassVar[Pattern[str]] = re.compile(r"\[[^\[\]]*\]")
 
-    _ONE_OR_TWO_DIGIT_RE: ClassVar[Pattern[str]] = re.compile(r"\d{1,2}")
+    _ONE_OR_TWO_DIGIT_RE: ClassVar[Pattern[str]] = re.compile(r"[1-9]\d|\d")
     _ONE_OR_TWO_OR_THREE_DIGIT_RE: ClassVar[Pattern[str]] = re.compile(r"\d{1,3}")
     _ONE_OR_MORE_DIGIT_RE: ClassVar[Pattern[str]] = re.compile(r"\d+")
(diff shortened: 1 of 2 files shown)


krisfremen (approved):
LGTM for fixing up the link issue.

As far as the 'DDD' token, ISO8601 IIRC specified ordinal numbers that can also have leading zeros. So I think it should be left as is.

TestArrowHumanize::test_months is marked as xfail because we were running into issues with the limits in specific dates/times. it's something we should definitely look again into in the future.

rodrigondec:
This PR is awaiting what? :cry:

Issue #617, opened 2019-07-17
Add Cython support to speed up arrow.get


jadchaar:
We had a PR in 2017 that addressed this issue (https://github.com/crsmithdev/arrow/pull/414), but it was never merged in, and the Cython implementation has since gotten outdated. It would be great if someone could take on this issue.

mashfy13:
I'm interested in helping out with the Cython implementation. I will definitely take a look at it!

mashfy13:
Is there any way to optimize the *args and **kwargs lists using Cython? I've been considering how to approach this but it seems like it would be pretty difficult since these lists can have different types based on what is getting passed to the function. I am currently thinking that these two lists would have to remain in their Python implementation.

jadchaar:
Hi @mashfy13, I am not too familiar with Cython myself, but we created this issue since the old PR became stale after a lack of review: https://github.com/crsmithdev/arrow/pull/414/. You can refer to the PR for a bit of inspiration. It may be best to focus the Cython optimizations on the parsing elements of Arrow since this may be used most often and in bulk.

I wish I could be of more help, but I have not gotten around to learning the ins and outs of Cython yet.

PR #1143, opened 2022-11-21
Cython

13MK3:
## Pull Request Checklist

## Description of Changes

These changes were made in collaboration with @khanm3

- Modified the setup file so that the entire Arrow library is compiled using Cython
- Modified the tox.ini file to support code coverage with Cython
- Added Cython 3.0.0a11 to the requirements.txt
- Modified arrow.py to compile with Cython and pass the test suite
- Modified an existing test case to work with Cython since the mocking was broken from Cython caching builtins
- Added a Makefile target to compile with Cython

This pull requests add Cython support to Arrow as requested by #617. 

## Further Explanation

In arrow/arrow.py, we had to change the variable "locale" to "local_cls" because the variable was being reused for 2 different types and this functionality was incompatible with Cython. Cython also had difficulties recognizing the type of _datetime.isoclanedar, so we had to specifically cast it to a tuple. Without this, it would not compile. 

Even with the issues mentioned above, we chose to use Cython 3.0.0a11 (which is a pre-release version) instead of Cython 0.29.32 (the most recent stable version), since from our testing, Cython 3 is better at parsing modern Python features. In general, Cython tries to subsume the functionality of python, but as you can see there are still many quirks.

We were able to allow Cython to use the existing method of calculating code coverage by adding linetracing as a compiler directive. However, due to the quirks of Cython, the resulting line coverage results are slightly lower than 100%. We found that the lines that were no longer being covered were one of 2 cases:
1) Inside an if statement that had 'pragma no cover' (so those also shouldn't count towards coverage)
2) Parameters to functions that had a default value (which also shouldn't count towards coverage, but for some reason Cython thinks it should)

This could be ignored by adding more 'pragma no cover', but we're not sure if there is a better way.

## Performance Analysis

To compare performance of Arrow with and without Cython, we ran just the Arrow part of [this benchmark](https://gist.github.com/freach/5490d8c0f86a0fff99e0b3e0d34205ad). This benchmark runs 1 million executions per benchmark and picked the min of 5 repeats. This was run in Python 3.8.10.

arrow_benchmark

From these results, you can see that compiling Arrow with Cython does cause a measureable improvement to performance, especially with parsing. Additionally, compiling Cython without linetracing is even more performant, but we currently compile with it so that code coverage can be calculated.

## Future Work

Adding static typing (and necessarily converting those files to .pyx files) would bring an even bigger boost to performance. Parser.py should receive special attention, as this is where the largest amount of time is being spend. This is where adding Cython to Arrow could prove to be especially valuable, since the speed increase with static typing should be much more than that gain by simply compiling with Cython. Looking into different ways of calculating code coverage so linetracing is no longer needed, or providing an option to switch the compiler directive would also be a good way to improve performance in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: auto test docs clean
+.PHONY: auto test docs clean compile-cython
 
 auto: build310
 
@@ -16,6 +16,10 @@ build36 build37 build38 build39 build310: clean
 	pip install -r requirements/requirements-tests.txt; \
 	pip install -r requirements/requirements-docs.txt; \
 	pre-commit install
+
+compile-cython:
+	. venv/bin/activate; \
+	python setup.py build_ext --inplace
 
 test:
 	rm -f .coverage coverage.xml
@@ -45,7 +49,7 @@ clean: clean-dist
 clean-dist:
 	rm -rf dist build .egg .eggs arrow.egg-info
 
-build-dist:
+build-dist: compile-cython
 	. venv/bin/activate; \
 	pip install -U pip setuptools twine wheel; \
 	python setup.py sdist bdist_wheel
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -1148,7 +1148,7 @@ class Arrow:
         """
 
         locale_name = locale
-        locale = locales.get_locale(locale)
+        locale_cls = locales.get_locale(locale)
 
         if other is None:
             utc = dt_datetime.utcnow().replace(tzinfo=dateutil_tz.tzutc())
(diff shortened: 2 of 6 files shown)


krisfremen:
Very nice, i had a wip branch of this, but this looks like good progress.

13MK3:
We realized the system tests are failing because the test systems aren't installing Cython before running setup.py. It seems that the continuous_integration.yml is using cached requirements and therefore not installing Cython even though we specified it in requirements.txt? We are not exactly sure and could use some help with this. Cython seems to install fine when we do "make build3x" locally.

krisfremen:
@13MK3 I see 3.8+ is passing, the rest are failing. Great work.

@anishnya @jadchaar @systemcatch how do we want to handle this going forward? do we want to go full cython or offer a pure python version as well?

PR #1179, opened 2024-07-07
Use zoneinfo for timezone handling instead of pytz

### LINKED ISSUE #1175: RFE: use `zoneinfo` instead of `pytz`
Switch to standard `zoneinfo` module.
Below may help  https://github.com/pydantic/pydantic-core/commit/fd262933

```console
[tkloczko@pers-jacek arrow-1.3.0]$ grep -r pytz
CHANGELOG.rst:- [FIX] Fixed a bug that occurred when ``arrow.Arrow()`` was instantiated with a ``pytz`` tzinfo object.
CHANGELOG.rst:- [FIX] Fix pytz conversion error (Kudo)
README.rst:- Too many modules: datetime, time, calendar, dateutil, pytz and more
README.rst:- Support for ``dateutil``, ``pytz``, and ``ZoneInfo`` tzinfo objects
arrow/arrow.py:        # detect that tzinfo is a pytz object (issue #626)
requirements/requirements-tests.txt:pytz==2021.1
tests/test_arrow.py:import pytz
tests/test_arrow.py:    def test_init_pytz_timezone(self):
tests/test_arrow.py:            2013, 2, 2, 12, 30, 45, 999999, tzinfo=pytz.timezone("Europe/Paris")
tests/test_formatter.py:import pytz
tests/test_formatter.py:        dt = datetime(1986, 2, 14, tzinfo=pytz.timezone("UTC")).replace(
tests/utils.py:import pytz
tests/utils.py:    pytz_zones = set(pytz.all_timezones)
tests/utils.py:    return dateutil_zones.union(pytz_zones)
pyproject.toml:    "pytz==2021.1",
```

saranti:
## Pull Request Checklist

- [ ] 🛠️  All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [ ] ⏩  Code is **up-to-date** with the `master` branch.

## Description of Changes

Closes #1175. Remove all pytz code and replace it with the equivalent from [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) and use a backport for python versions <3.9.


pyproject.toml
```diff
@@ -46,7 +46,6 @@ test = [
     "pytest",
     "pytest-cov",
     "pytest-mock",
-    "pytz==2021.1",
     "simplejson==3.*",
```
jadchaar, line 49: we need to add a dependency on the backport like this:
```
backports.zoneinfo = {version = "0.2.1", markers = "python_version < '3.9'"}
```

tests/utils.py
```diff
@@ -1,11 +1,14 @@
...
+except ImportError:
+    from backports import zoneinfo
 from dateutil.zoneinfo import get_zonefile_instance
 
 
 def make_full_tz_list():
     dateutil_zones = set(get_zonefile_instance().zones)
-    pytz_zones = set(pytz.all_timezones)
-    return dateutil_zones.union(pytz_zones)
+    zoneinfo_zones = set(zoneinfo.available_timezones())
```
jadchaar, line 10: Does this have the proper coverage for all timezones? We primarily used pytz to get greater coverage of time zones for the sake of tests

tests/test_arrow.py
```diff
@@ -57,9 +56,9 @@ def test_init(self):
         assert result._datetime == self.expected
 
     # regression tests for issue #626
-    def test_init_pytz_timezone(self):
```
jadchaar, line 60: Can we add a separate test for pytz and another for zoneinfo? I don't think we should be removing pytz and closing the open issue for the sake of the tests. There is still a use-case for passing in a pytz timezone into Arrow, especially if someone has not yet adopted zoneinfo.

If pytz is not used in the core arrow code and only in the test code (it is only pulled in via `requirements-tests.txt`, I think we need another CR that focuses on adding support for Zoneinfo natively throughout rather than using dateutil, which is our primary means of doing timezones.

jadchaar (changes requested):

tests/utils.py
saranti: I just compared them and it does have the same coverage as pytz (Windows 11). However, from the [zoneinfo docs](https://docs.python.org/3/library/zoneinfo.html#data-sources):

> The zoneinfo module does not directly provide time zone data, and instead pulls time zone information from the system time zone database or the first-party PyPI package [tzdata](https://pypi.org/project/tzdata/), if available. Some systems, including notably Windows systems, do not have an IANA database available, and so for projects targeting cross-platform compatibility that require time zone data, it is recommended to declare a dependency on tzdata. If neither system data nor tzdata are available, all calls to [ZoneInfo](https://docs.python.org/3/library/zoneinfo.html#zoneinfo.ZoneInfo) will raise [ZoneInfoNotFoundError](https://docs.python.org/3/library/zoneinfo.html#zoneinfo.ZoneInfoNotFoundError).

tests/test_arrow.py
saranti: Sure

jadchaar:
Hi @saranti mind getting this PR up to date? Then we can merge it in :)

jadchaar:
Thanks @saranti , but it seems like the conflicts are still present and GH is not allowing us to re-run the actions to test

saranti:
No worries. That looks better but it looks like there's a linting issue caused by https://github.com/arrow-py/arrow/commit/c2dfa12bde6bf8b097ad5bb369a0c71ae58a8386

krisfremen:
> No worries. That looks better but it looks like there's a linting issue caused by [c2dfa12](https://github.com/arrow-py/arrow/commit/c2dfa12bde6bf8b097ad5bb369a0c71ae58a8386)

Strange indeed, we might need to look at the linting actions and see why this got missed in the first place before merging.

https://github.com/arrow-py/arrow/pull/1202 Got a PR to revert that one @jadchaar

jadchaar:
Odd indeed, this must have come as a result of the upgrades we made to the actions. I merged the revert, and updated branch @krisfremen. Would it be possible to reapply the PR you reverted once we merge?

PR #1192, opened 2024-10-07
Fix daylight saving time issue in shift function

### LINKED ISSUE #1170: incorrect behavior WRT daylight savings for shift/adding timedelta
## Issue Description
Arrow's `shift` function appears to handle daylight savings time incorrectly. In particular, on spring-forward days, shifting forward by 2 hours from midnight is treated as being the same time as shifting 3 from midnight:
```
Python 3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:41:52) [Clang 15.0.7 ] on darwin
>>> import arrow
>>> x = arrow.get(datetime(2023, 3, 12), 'America/Los_Angeles')
>>> x.shift(hours=3) == x.shift(hours=2)
True
>>> print(x.shift(hours=2))
2023-03-12T03:00:00-07:00
>>> print(x.shift(hours=3))
2023-03-12T03:00:00-07:00
```
It looks to me like shifting by 2 hours is correct, but that shifting 3 is not. Interestingly, shifting by 1 hour three times does what I'd expect:
```
>>> x.shift(hours=1).shift(hours=1).shift(hours=1)
<Arrow [2023-03-12T04:00:00-07:00]>
>>> x.shift(hours=1).shift(hours=1).shift(hours=1) == x.shift(hours=3) # I'd expect this to be true for any value of x
False
```

When I try adding `timedelta(hours=1)` I  get different unintuitive behavior, this time when it comes to adding 1 hour or 2 hours:
```
>>> (x + timedelta(hours=1)).timestamp() == (x + timedelta(hours=2)).timestamp() # I'd expect this to be false
True
>>> x + timedelta(hours=1)
<Arrow [2023-03-12T01:00:00-08:00]> # this looks right to me
>>> x + timedelta(hours=2)
<Arrow [2023-03-12T02:00:00-07:00]> # I think this should be 2023-03-12T03:00:00-07:00

```

On fall-behind days, there are also problems. In this case using both `shift` and adding a `timedelta` do the same thing : the gap between midnight offsetby 1 hour and offset by 2 hours is 7200 seconds (2 hours):
```
>>> y = arrow.get(datetime(2023, 11, 5), 'America/Los_Angeles')
>>> y.shift(hours=1)
<Arrow [2023-11-05T01:00:00-07:00]>
>>> y.shift(hours=2)
<Arrow [2023-11-05T02:00:00-08:00]>
>>> y.shift(hours=2).timestamp() - y.shift(hours=1).timestamp() # I'd expect 3600
7200.0
>>> (y + timedelta(hours=2)).timestamp() - (y + timedelta(hours=1)).timestamp() # I'd expect 3600
7200.0
```

## System Info

- 🖥  **OS name and version**: macOs 14.2.1
- 🐍  **Python version**: 3.10.12 
- 🏹  **Arrow version**: 1.3.0

Shaileshsaravanan:
## Pull Request Checklist

### Fixes #1170: Correct DST Handling in `shift` Function

This PR addresses the issue with incorrect daylight saving time (DST) handling when using the `shift` function in the Arrow library.

#### Changes:
- Modified the `shift` function to correctly handle transitions during DST changes.
- Added logic to detect positive or negative offset changes when shifting across DST boundaries.
- Adjusted the behavior during spring-forward (to skip the missing hour) and fall-back (to use the second occurrence of a repeated hour).

#### Issue Link:
Closes: #1170

#### Explanation:
Previously, the `shift` function did not account for the changes in offsets that occur during DST transitions, leading to unexpected behavior when shifting times. This fix ensures that:
- For spring-forward transitions, any shifts that would land in the missing hour are adjusted forward.
- For fall-back transitions, the repeated hour is handled properly by using the `fold` attribute to select the second occurrence.

#### Testing:
- Added tests for time shifts that cross DST start and end dates to verify the behavior.
- Ensured consistency when using `timedelta` shifts and repeated shifts.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -1029,12 +1029,35 @@ class Arrow:
             relative_kwargs.pop("quarters", 0) * self._MONTHS_PER_QUARTER
         )
 
+        # Store the original datetime before shifting
+        original_datetime = self._datetime
+
+        # Shift using relative delta
         current = self._datetime + relativedelta(**relative_kwargs)
 
+        # Check if the shifted datetime exists in the timezone
         if not dateutil_tz.datetime_exists(current):
             current = dateutil_tz.resolve_imaginary(current)
 
+        # Adjust for DST transitions if there's a change in offset
+        original_offset = original_datetime.utcoffset()
+        new_offset = current.utcoffset()
+
+        if new_offset != original_offset:
+            offset_difference = new_offset - original_offset
+
+            # Adjust for the spring-forward transition (skipped hour)
+            if offset_difference.total_seconds() > 0:
+                # Add the offset difference to the current time
+                current += offset_difference
+
+            # Adjust for the fall-back transition (repeated hour)
+            elif offset_difference.total_seconds() < 0:
+                # Use `fold=1` to indicate the second occurrence of the repeated hour
+                current = current.replace(fold=1)
+
         return self.fromdatetime(current)
+
 
     def to(self, tz: TZ_EXPR) -> "Arrow":
         """Returns a new :class:`Arrow <arrow.arrow.Arrow>` object, converted


jadchaar:
Hi @Shaileshsaravanan thanks for the PR. Mind getting the builds and tests passing? Then we can review :)

satveerbrar:
@jadchaar 
Arrow uses the `resolve_imaginary` function from `dateutil` library to handle DST offset. That seems to be the problem here because it only fix the offset if the time is invalid. For example, on `2023-03-12T02:00:00-07:00` in `America/Los_Angeles`, this time does not exist due to the DST change.  In this case `resolve_imaginary` will correct the time. But if shift function directly shift the time to valid time then this function not going to make any changes

Look into the code for reference
https://github.com/dateutil/dateutil/blob/9eaa5de584f9f374c6e4943069925cc53522ad61/src/dateutil/tz/tz.py#L1763

Issue #1145, opened 2022-11-30
passing a parameter to Arrow.shift function (to check for imaginary times or not)


not-so-rabh:
The shift function takes some time than absolute adding of relativedelta to datetime. If I am operating on a range where I know there is no dst change I should be able to bypass this check.

PR #1194, opened 2024-10-17
Update shift() for issue #1145

psyuktha:
## Pull Request Checklist

## Description of Changes

Added check_imaginary parameter to the function referring to the issue #1145 
- New Parameter: Added a check_imaginary parameter (defaulting to True), which allows the user to decide whether or not to perform DST/imaginary time checks.
- Skip DST Check: The DST check logic:
block is only executed if check_imaginary is True. If set to False, this check is skipped for improved performance when DST changes are not a concern.

Benefits:
- Backward Compatibility: By default, the method still performs the DST/imaginary time checks, so existing code won't break.
- Performance Boost: Users who know that their datetime range does not involve DST transitions can set check_imaginary=False to skip the checks and improve performance.

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -491,7 +491,7 @@ class Arrow:
             yield current
 
             values = [getattr(current, f) for f in cls._ATTRS]
-            current = cls(*values).shift(
+            current = cls(*values, tzinfo=tzinfo).shift(  # type: ignore[misc]
                 check_imaginary=True, **{frame_relative: relative_steps}
             )
 
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -491,8 +491,8 @@ class Arrow:
             yield current
 
             values = [getattr(current, f) for f in cls._ATTRS]
-            current = cls(*values, tzinfo=tzinfo).shift(  # type: ignore[misc]
-                **{frame_relative: relative_steps}
+            current = cls(*values).shift(
+                check_imaginary=True, **{frame_relative: relative_steps}
             )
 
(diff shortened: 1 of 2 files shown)


arrow/arrow.py
```diff
@@ -492,7 +492,7 @@ def range(
 
             values = [getattr(current, f) for f in cls._ATTRS]
             current = cls(*values, tzinfo=tzinfo).shift(  # type: ignore[misc]
-                **{frame_relative: relative_steps}
+                check_imaginary=True, **{frame_relative: relative_steps}
```
jadchaar, line 495: Mind explaining why we want to force usage of check imaginary here rather than making it an argument?

psyuktha: To bypass unnecessary checks when performance matters and they know they won't encounter DST transitions.

The default behavior of check_imaginary=True ensures that users get correct and valid results by default. If a user deliberately wants to skip the check for performance reasons or because they're confident it isn't needed, they can do so by setting the parameter to False.

jadchaar: Ah my bad, I now see that this was added as true in range and dehumanize shift usages to maintain backwards compatibility right?

psyuktha: Yes exactly

psyuktha:
@jadchaar If all the requirements are met, can you please approve it?

jadchaar (approved):
LGTM

psyuktha:
Could you please add a Hacktoberfest accepted label to this, will be thankful

PR #1242, opened 2026-02-08
Fix humanize reporting 'a month' for 16-day differences

### LINKED ISSUE #1240: Strange `humanize` intervals
## Issue Description

It seems that the intervals being used for the various thresholds for `humanize` are a bit strange. For example, on one of my websites, I have a list of upcoming events with both the calendar date and a `humanize`d description. As of today, January 9 2026, there are two events, one on January 24 which appears as "in two weeks" and one on January 25 which appears as "in a month."

15 days (the 24th) is indeed about two weeks away, but 16 days (the 25th) is certainly not a month away - if anything it's also closest to around two weeks away. It isn't even closer to three weeks than two, less alone a month.

## System Info

- 🖥  **OS name and version**: Linux (Ubuntu 24.04 LTS if that matters)
- 🐍  **Python version**: 3.12.3
- 🏹  **Arrow version**: 1.4.0

veeceey:
## Summary

Fixes #1240.

- **Bug**: `humanize` reported 16 days in the future as "in a month" instead of "in 2 weeks" because `calendar_diff.days > 14` unconditionally bumped `calendar_months` from 0 to 1, even when no calendar month boundary was crossed.
- **Fix**: Only round up partial months when `calendar_months >= 1` (an actual month boundary was crossed), and check weeks *before* months (guarded by `calendar_months < 1`) so within-month differences use the weeks granularity while true calendar-month differences (e.g., Feb 8 → Mar 8 = 28 days) still correctly show as "a month."

### Before (broken)
| Delta | Result |
|-------|--------|
| +15 days (same month) | "in 2 weeks" |
| +16 days (same month) | "in a month" |

### After (fixed)
| Delta | Result |
|-------|--------|
| +15 days (same month) | "in 2 weeks" |
| +16 days (same month) | "in 2 weeks" |
| +28 days (Feb→Mar, shift(months=1)) | "in a month" |
| +31 days (crosses month boundary) | "in a month" |

## Test plan

- [x] All 225 existing tests pass (including all 50 humanize tests)
- [x] Verified the exact scenario from the issue: Jan 9 → Jan 25 (16 days) now shows "in 2 weeks" instead of "in a month"
- [x] Verified short-month edge case: `shift(months=1)` from February (28 days) correctly shows "in a month"
- [x] Verified smooth progression: days → week → weeks → month → months

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -1223,8 +1223,10 @@ class Arrow:
                     calendar_diff.years * self._MONTHS_PER_YEAR + calendar_diff.months
                 )
 
-                # For months, if more than 2 weeks, count as a full month
-                if calendar_diff.days > 14:
+                # Round up partial months when already at least one
+                # calendar month has elapsed and the remaining days
+                # are more than halfway through another month.
+                if calendar_months >= 1 and calendar_diff.days > 14:
                     calendar_months += 1
 
                 calendar_months = min(calendar_months, self._MONTHS_PER_YEAR)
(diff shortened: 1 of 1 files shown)


krisfremen:
excellent, I was just looking at fixing this up yesterday, I will give this a review.

veeceey:
Thanks! Appreciate you taking a look. Let me know if anything needs changing.

krisfremen (approved):

PR #1243, opened 2026-02-13
fix IndexError in DateTimeParser.parse with malformed format strings

### LINKED ISSUE #1191: IndexError in DateTimeParser.parse
```
import sys

parser = DateTimeParser()
parser.parse("foo", buf)
```
### Version
1.3.0

### POC File

### How to trigger
python filename.py poc-01

### Trace Report
```
    parser.parse("foo", buf)
    value = match.group(token)
IndexError: no such group
```

Mr-Neutr0n:
Malformed format strings can cause `DateTimeParser.parse()` to raise an `IndexError` instead of a `ParserMatchError`. This happens when the format string contains characters that, after regex escaping and bracket expression handling, produce a token list that doesn't align with the named groups in the generated regex pattern. `match.group(token)` then blows up with `IndexError: no such group`.

The fix wraps the `match.group()` calls in a try/except to catch `IndexError` and raise `ParserMatchError` instead, which is the expected exception type for parse failures.

Reproducer from #1191:
```python
p = DateTimeParser()
p.parse("foo", "[-FFFFFFFFFFFFFFFFFFFFF-[-FFFFFFFFFFFF[ |||||||| 7   v.dG(dG\\][3zasks  &")
# Before: IndexError: no such group
# After: ParserMatchError: Failed to match '...' when parsing 'foo'.
```

Added a regression test. All 723 existing parser tests still pass.

Fixes #1191


Mr-Neutr0n:
Hi! Quick status: this still applies cleanly to the latest main as of today, and the PR is still mergeable. Re-requesting a review. If this looks good now, I would be happy to see it merged; if the repo has moved on, I am just as happy to close it.

Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

PR #1256, opened 2026-02-26
Fix MAX_TIMESTAMP calculation to use UTC timezone

### LINKED ISSUE #1095: MAX_TIMESTAMP calculation error
## Issue Description

There are some issues related to the way max values are handled (partially because of how python's datetime works) : 

The way how Arrow calculates the max timestamp constant value also fails on Linux in **constants.py** : 

`_MAX_TIMESTAMP = datetime.max.timestamp()`

Which will raise a ValueError in non-utc timezones (timestamp() ) will try to return a timestamp in the local timezone):

`ValueError: year 10000 is out of range`

Therefore I suggest making this more robust and replacing this with : 

`datetime.max.replace(tzinfo=timezone.utc).timestamp()`

Which is more in line with python documentation : https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp

Using `Arrow.max` has its own issues , as

`Arrow.utcfromtimestamp(Arrow.max.float_timestamp)`

does not behave as expected and returns a date in 1978, due to  `normalize_timestamp`  (util) division by 1000 : 

```python
def normalize_timestamp(timestamp: float) -> float:
    """Normalize millisecond and microsecond timestamps into normal timestamps."""
    if timestamp > MAX_TIMESTAMP:
        if timestamp < MAX_TIMESTAMP_MS:
            timestamp /= 1000
        elif timestamp < MAX_TIMESTAMP_US:
            timestamp /= 1_000_000
        else:
            raise ValueError(f"The specified timestamp {timestamp!r} is too large.")
    return timestamp
```

## System Info

- 🖥  **OS name and version**: Debian 11
- 🐍  **Python version**: 3.9.2
- 🏹  **Arrow version**: 1.2.2

### LINKED ISSUE #1159: error in Time transform of 9999-12-31
## Issue Description
i want to change '9999-12-31' to timestamp, and also in anather scene, i need change the timestamp(253402214400) to datetime.
but I got confused when I test this :
```
 dt2=arrow.get('9999-12-31')
 dt2.timestamp()
```
then get:
> 253402214400.0

and change it to datetime:
`arrow.get(253402214400.0)`

but get  :
> <Arrow [1978-01-11T21:30:14.400000+00:00]>

this code should get result like of : "9999-12-31 "

So, It is a bug , or my code is not correct！

## System Info

- 🖥  **OS name and version**: winodws11 and wsl ubuntu
- 🐍  **Python version**: python3.7
- 🏹  **Arrow version**: 1.2.3

rstar327:
## Pull Request Checklist

## Description of Changes

Closes: #1095
Closes: #1159

`MAX_TIMESTAMP` in `arrow/constants.py` is calculated using `datetime.max.timestamp()`, which interprets the naive datetime as local time. This means the resulting timestamp varies depending on the system's local timezone, leading to inconsistent behavior across different environments.

This PR fixes the issue by explicitly setting the timezone to UTC before calling `.timestamp()`:

```python
# Before (timezone-dependent)
_MAX_TIMESTAMP = datetime.max.timestamp()

# After (always UTC)
_MAX_TIMESTAMP = datetime.max.replace(tzinfo=timezone.utc).timestamp()
```

The same fix is applied to the Windows/32-bit fallback values to ensure consistency across all platforms.

**Tests added:**
- `test_max_timestamp_is_utc_based` — verifies `MAX_TIMESTAMP` equals the UTC-based calculation
- `test_max_timestamp_derived_constants` — verifies `MAX_TIMESTAMP_MS` and `MAX_TIMESTAMP_US` are correctly derived

All 1905 tests pass with 99.93% coverage.

### DIFF
--- a/arrow/constants.py
+++ b/arrow/constants.py
@@ -1,7 +1,7 @@
 """Constants used internally in arrow."""
 
 import sys
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Final
 
 # datetime.max.timestamp() errors on Windows, so we must hardcode
@@ -11,7 +11,7 @@ from typing import Final
 try:
     # Get max timestamp. Works on POSIX-based systems like Linux and macOS,
     # but will trigger an OverflowError, ValueError, or OSError on Windows
-    _MAX_TIMESTAMP = datetime.max.timestamp()
+    _MAX_TIMESTAMP = datetime.max.replace(tzinfo=timezone.utc).timestamp()
 except (OverflowError, ValueError, OSError):  # pragma: no cover
     # Fallback for Windows and 32-bit systems if initial max timestamp call fails
     # Must get max value of ctime on Windows based on architecture (x32 vs x64)
@@ -19,9 +19,9 @@ except (OverflowError, ValueError, OSError):  # pragma: no cover
     # Note: this may occur on both 32-bit Linux systems (issue #930) along with Windows systems
     is_64bits = sys.maxsize > 2**32
     _MAX_TIMESTAMP = (
-        datetime(3000, 1, 1, 23, 59, 59, 999999).timestamp()
+        datetime(3000, 1, 1, 23, 59, 59, 999999, tzinfo=timezone.utc).timestamp()
         if is_64bits
-        else datetime(2038, 1, 1, 23, 59, 59, 999999).timestamp()
+        else datetime(2038, 1, 1, 23, 59, 59, 999999, tzinfo=timezone.utc).timestamp()
     )
 
 MAX_TIMESTAMP: Final[float] = _MAX_TIMESTAMP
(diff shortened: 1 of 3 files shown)


daphne-barretto:
+1, this also should resolve #1159

here’s a regression test for #1159 that can be added to tests/test_util.py to validate this moving forward:

```python
# GH #1159: a valid second-precision timestamp near datetime.max must be
# left untouched, regardless of the host's local time zone. Previously the
# constant collapsed to a year-3000 fallback on non-UTC hosts, so this value
# exceeded MAX_TIMESTAMP and was wrongly rescaled as ms -> a 1978 datetime.
max_year_timestamp = 253402214400.0  # arrow.get("9999-12-31").timestamp()
assert max_year_timestamp <= constants.MAX_TIMESTAMP
assert util.normalize_timestamp(max_year_timestamp) == max_year_timestamp
```

(requires `from arrow import constants, util` at the top of the file.)

PR #1261, opened 2026-03-14
Add fold kwarg to arrow.get() for DST disambiguation

### LINKED ISSUE #872: Allow fold kwarg in arrow.get()
Currently there is no way to specify which side of an ambiguous datetime we want when creating with `arrow.get()`. It'd be nice to be able to do the following.

```python
arrow.get("2001-10-28 01:00:00", "YYYY-MM-DD HH:mm:ss", tzinfo="US/Pacific", fold=1)
```

wavebyrd:
## Summary
- Adds support for the `fold` parameter in `arrow.get()` to allow users to disambiguate times that occur twice during DST transitions
- The `fold` parameter works when parsing ISO strings or format strings
- Includes tests for the new functionality

## Example Usage
```python
import arrow

# During DST "fall back", 1:30 AM occurs twice
# fold=0 selects the first (daylight time), fold=1 selects the second (standard time)
arrow.get("2001-10-28T01:30:00", tzinfo="US/Pacific", fold=0)  # First occurrence
arrow.get("2001-10-28T01:30:00", tzinfo="US/Pacific", fold=1)  # Second occurrence
```

## Test plan
- [x] Added tests for fold kwarg with ISO string parsing
- [x] Added tests for fold kwarg with format string parsing
- [x] Added test verifying default behavior when fold is not specified
- [x] All new tests pass locally

Fixes #872


wavebyrd:
Friendly ping - this fixes type hints to accept Arrow objects in interval()/span_range() functions. Let me know if any changes needed!

PR #1262, opened 2026-03-14
Fix type hints for interval and span_range to accept Arrow objects

### LINKED ISSUE #1210: Type hints for `Arrow.interval` should show `Arrow` as possible type for `start` and `end` arguments
## Issue Description

Currently, footprint for `Arrow.interval` is:
```py
    def interval(
        cls,
        frame: _T_FRAMES,
        start: dt_datetime,
        end: dt_datetime,
        interval: int = 1,
        tz: Optional[TZ_EXPR] = None,
        bounds: _BOUNDS = "[)",
        exact: bool = False,
    ) -> Iterable[Tuple["Arrow", "Arrow"]]:
```
Despite the fact that `start` and `end` arguments have type `dt_datetime` (which is `datetime.datetime`) however, both of them accept `Arrow` objects, which, looking down the line, makes sense, as the only properties of these arguments that are used are shared between `Arrow` and `dt_datetime` and (see `fromdatetime` method).
Thus, i believe it would make sense to change type hints for both arguments to the union of `Arrow` and `dt_datetime`: 
```py
    def interval(
        cls,
        frame: _T_FRAMES,
        start: "Arrow" | dt_datetime,
        end: "Arrow" | dt_datetime,
        interval: int = 1,
        tz: Optional[TZ_EXPR] = None,
        bounds: _BOUNDS = "[)",
        exact: bool = False,
    ) -> Iterable[Tuple["Arrow", "Arrow"]]:
```

## System Info

- 🖥  **OS name and version**: Windows 24H2 26100.3476
- 🐍  **Python version**: 3.10.11
- 🏹  **Arrow version**: 1.3.0

wavebyrd:
## Summary
- Updates the type hints for `start` and `end` parameters in `Arrow.interval()` and `Arrow.span_range()` to `Union["Arrow", dt_datetime]`
- This makes the type hints match the documented behavior and be consistent with `Arrow.range()` which already has the correct type hints

## Changes
- `Arrow.interval()`: Changed `start` and `end` from `dt_datetime` to `Union["Arrow", dt_datetime]`
- `Arrow.span_range()`: Changed `start` and `end` from `dt_datetime` to `Union["Arrow", dt_datetime]`

## Test plan
- [x] Module imports correctly
- [x] Type hints now accurately reflect the documented API behavior

Fixes #1210


wavebyrd:
Ping on this fold kwarg addition for DST disambiguation. Let me know if any changes needed!

Issue #1259, opened 2026-03-05
arrow.get() behaviour for tzinfo=None


suzukieng:
## Issue Description

When `arrow.get()` is invoked with a value, a format and a `tzinfo` keyword argument that is `None`, the behavior is different than if the keyword argument were omitted, which is slightly confusing (at least to me).

Example:

```python
d = arrow.get('2025-01-01', 'YYYY-MM-DD', tzinfo=account.timezone)
```
where account.timezone is optional and can be None.

This will crash because it will attempt to invoke the 3+ argument constructor of `Arrow`:
```
        # 3+ args -> datetime-like via constructor
        else:
>           return self.type(*args, **kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
E           TypeError: Arrow.__init__() missing 1 required positional argument: 'day'
```

The reason is that in the factory, the check assumes that a single keyword argument and a `None` tzinfo means we've not passed a tzinfo, when in fact we've passed `None`. 

```
        # tzinfo kwarg is not provided
        if len(kwargs) == 1 and tz is None:
            arg_count = 3
```

Personally I think explicitly passing `None` as tzinfo should cause it be to treated as 'not passed at all', but I'm not a Python expert, so feel free to close as "works as intended".

## System Info

- 🖥  not relevant
- 🐍  3.13
- 🏹  1.3.0

PR #1263, opened 2026-03-17
fix: handle tzinfo=None kwarg in factory.get() without crashing

frankgoldfish:
## Summary

Fixes #1259.

When `tzinfo=None` is explicitly passed as a keyword argument to `arrow.get()`, the factory crashed with:

```
TypeError: __init__() missing 1 required positional argument: 'day'
```

## Root Cause

In `arrow/factory.py`, the check:

```python
if len(kwargs) == 1 and tz is None:
    arg_count = 3
```

incorrectly treats `tzinfo=None` (explicitly passed) as equivalent to "no tzinfo was passed at all", which forces `arg_count = 3` and routes into the positional year/month/day constructor branch — even though no year/month/day args were given.

## Fix

Change `tz is None` to `"tzinfo" not in kwargs`:

```python
    arg_count = 3
```

This correctly distinguishes between:
- `tzinfo` kwarg absent → old behavior preserved  
- `tzinfo=None` explicitly passed → treated as a valid tzinfo value (defaults to UTC)

## Reproduction

```python
import arrow

# Before fix: TypeError
# After fix: works correctly
arrow.get("2025-01-01", "YYYY-MM-DD", tzinfo=None)
# Returns: <Arrow [2025-01-01T00:00:00+00:00]>
```

Real-world pattern: `tzinfo=account.timezone` where `account.timezone` is an optional field that may be `None`.

## Tests

Added `test_tzinfo_none_kwargs` regression test in `tests/test_factory.py`. All 49 tests pass.


frankgoldfish:
Closing this PR — found that [PR #1260](https://github.com/arrow-py/arrow/pull/1260) already addresses this issue. Apologies for the duplicate.

PR #1267, opened 2026-04-03
fix(types): use Union[Arrow, dt_datetime] for span_range and interval start/end params

### LINKED ISSUE #1210: Type hints for `Arrow.interval` should show `Arrow` as possible type for `start` and `end` arguments

juliosuas:
Fixes #1210.

`Arrow.span_range()` and `Arrow.interval()` accept `start` and `end` bounds like
`Arrow.range()` does, but their annotations only allowed `datetime`.

This updates those annotations to accept `Arrow | datetime` and normalizes
`Arrow` bounds before passing them through `fromdatetime()`.

The test dependency markers also skip two optional test-only dependencies on
Windows PyPy. The affected `dateparser` regression test already uses
`pytest.importorskip("dateparser")`, and `pre-commit` is only needed by the
lint environment.

Tests:

```bash
tox -e lint
tox -e py314
```

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -643,8 +643,8 @@ class Arrow:
     def span_range(
         cls,
         frame: _T_FRAMES,
-        start: dt_datetime,
-        end: dt_datetime,
+        start: Union["Arrow", dt_datetime],
+        end: Union["Arrow", dt_datetime],
         tz: Optional[TZ_EXPR] = None,
         limit: Optional[int] = None,
         bounds: _BOUNDS = "[)",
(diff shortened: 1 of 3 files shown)


juliosuas:
😁

juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/arrow-py/arrow/pull/1342

PR #1289, opened 2026-06-07
Fix get() crashing when tzinfo=None is passed explicitly

vincere-mori:
## Issue

`arrow.get()` crashes when `tzinfo=None` is passed explicitly alongside a value (and optionally a format), even though omitting `tzinfo` works fine. It comes up when the timezone is taken from an optional field:

```python
arrow.get("2025-01-01", "YYYY-MM-DD", tzinfo=account.timezone)  # account.timezone may be None
```

```
TypeError: Arrow.__init__() missing 1 required positional argument: 'day'
```

Reported in #1259.

## Cause

`get()` decides whether a call is constructor-style (`get(2013, 5, 5, ...)`) or value-style by counting args, with this branch:

```python
# tzinfo kwarg is not provided
if len(kwargs) == 1 and tz is None:
    arg_count = 3
```

`tz is None` is used as a proxy for "no tzinfo was given", but it's also true when `tzinfo=None` is passed explicitly, so the call gets routed to the `Arrow` constructor and fails.

## Fix

Check whether the `tzinfo` key is actually present instead of relying on its value, which is what the comment already describes. An explicit `tzinfo=None` then behaves the same as omitting it (UTC):

```python
    arg_count = 3
```

Constructor dispatch for a real non-tzinfo kwarg (e.g. `get(2013, 5, 5, hour=12)`) is unchanged.

## Tests

Added a regression test for both the one-arg and two-arg forms with `tzinfo=None`. It fails on the current code with the `TypeError` above and passes with the fix; the rest of the suite stays green.

Happy to add a CHANGELOG entry if you'd like one.


vincere-mori:
The `windows-latest (pypy-3.11)` job failed while pip-installing the tox deps (`Connection broken: IncompleteRead`), so the tests never ran - the other 26 jobs are green. Could you re-run that one job when you get a chance? Thanks!

Issue #1278, opened 2026-05-17
dehumanize() silently ignores negative time values, returning wrong results


amanv0007:
### What happened

While using `dehumanize()` to parse user-provided relative time strings, I noticed that negative values like "in -1 hours" don't raise an error - they silently produce the **same result** as "in 1 hours". The minus sign is dropped without any warning.

This is because the internal number-extraction regex `\d+` only matches unsigned digits, so the `-` in `-1` is never captured. The result is a quietly wrong datetime that's in the opposite direction from what a user might expect.

### How to reproduce

```python
import arrow

now = arrow.now()

# These two return the exact same result (+1 hour):
result1 = now.dehumanize("in 1 hours")
result2 = now.dehumanize("in -1 hours")

print((result1 - now).total_seconds())  # 3600.0
print((result2 - now).total_seconds())  # 3600.0  <- should not be the same!
```

Similarly, "in -2 days" silently becomes +2 days, and "-3 minutes ago" silently becomes -3 minutes (dropping the sign).

### Expected behavior

`dehumanize()` should raise a `ValueError` when the input contains a negative number, since humanized time strings use "ago"/"in" for direction - not arithmetic signs. Silently swallowing the sign leads to wrong results that are hard to debug.

### Version

```
arrow 1.4.0
Python 3.12
Windows 11
```

PR #1292, opened 2026-06-08
Raise ValueError in dehumanize() for negative numbers (#1278)

vincere-mori:
Closes #1278.

## Issue

`dehumanize()` silently drops a minus sign on the number, so a negative value parses the same as the positive one:

```python
now = arrow.now()
(now.dehumanize("in 1 hours")  - now).total_seconds()   # 3600
(now.dehumanize("in -1 hours") - now).total_seconds()   # 3600  <- sign ignored
```

The internal number regex is `\d+`, which never captures the `-`, so the value comes out positive and the result is quietly wrong.

## Fix

Direction in a humanized string is expressed by the locale's past/future wording (`"ago"` / `"in"`), not by a sign on the number, so a negative number is invalid input. `dehumanize()` already raises `ValueError` for other invalid input (unknown units, missing relative unit, etc.), so this rejects negative numbers the same way instead of returning a silently wrong datetime:

```python
>>> arrow.now().dehumanize("in -1 hours")
ValueError: Invalid input string: negative numbers are not supported. ...
```

## Tests

Added `TestArrowDehumanize::test_negative_number` covering `"in -1 hours"` and `"-2 days ago"`. It fails on `master` (no error raised) and passes with the fix; the rest of the suite stays green and `ruff` is clean.


vincere-mori:
The `windows-latest (pypy-3.11)` job failed while pip-installing the tox deps (`Connection broken: IncompleteRead`), so the tests didn't run - the other 26 jobs are green. A re-run of that job should clear it.

PR #1298, opened 2026-06-18
fix: reject invalid trailing characters and unclosed UTC parens in TzinfoParser

### LINKED ISSUE #1294: Timezone parsing accepts invalid values
## Issue Description

This is accepted as a valid timezone expression, and should not be: `+03:00.34`

Note that this is a regression, and this PR would fix it: https://github.com/arrow-py/arrow/pull/1165

## System Info

- 🖥  **OS name and version**: Ubuntu 25.10
- 🐍  **Python version**: Python 3.13.7
- 🏹  **Arrow version**: 1.4.0

gaoflow:
Fixes #1294. The TzinfoParser regex accepted inputs like `+03:00.34`, `+03:00 `, and `(UTC+01:00` (missing closing paren) because it did not require the match to consume the entire string or properly close the UTC parenthetical.

**Change:** Updated the regex to use a conditional group — when the UTC parenthetical is present, a closing `)` is required; otherwise `$` (end-of-string) is required. The unpacking gains a dummy `_` to absorb the new capture group.

**Tests:** Extended `test_parse_fails` with the invalid inputs from the issue. Adjusted `test_parse_utc_withoffset` to use properly closed `(UTC±HH:MM)` forms.

This adapts the fix originally proposed in #1165 (open since 2023).


binggao1230:
CI triage note: the only red job, `windows-latest (pypy-3.11)`, failed before tests ran while installing dependencies:

`Connection broken: IncompleteRead(...)`

All other Python/platform jobs, linting, and Codecov are green, so this looks like a transient network/package download failure rather than a code failure. I attempted to rerun failed jobs, but GitHub requires repository admin rights for this workflow rerun.

PR #1310, opened 2026-07-04
Reject negative numbers in dehumanize()

eeshsaxena:
## Summary

Fixes #1278.

`dehumanize()` extracts numbers with an unsigned `\d+` pattern, and the surrounding match anchors on a word boundary, so a minus sign attached to a number (e.g. `"in -1 hours"`) is silently discarded. The result is a datetime in the **opposite** direction from what the string expressed.

### Before
```python
>>> now = arrow.now()
>>> (now.dehumanize("in 1 hours")  - now).total_seconds()
3600.0
>>> (now.dehumanize("in -1 hours") - now).total_seconds()
3600.0   # <- same as +1, the minus was dropped
```

### After
```python
>>> now.dehumanize("in -1 hours")
ValueError: Invalid input string: 'in -1 hours'. Negative numbers are not supported; use 'ago' or 'in' to indicate direction.
```

## Rationale

Humanized strings encode direction with words (`"ago"`/`"in"`), never with an arithmetic sign, so `humanize()` itself never emits one. A negative number in the input is therefore always malformed, and returning a quietly-wrong result is worse than raising. The guard rejects any `-<digit>` in the input; all existing positive/`ago`/`in`/`now` inputs are unaffected.

## Tests

Added `test_negative_number` (covers `"in -1 hours"`, `"in -2 days"`, `"-3 minutes ago"`). Full dehumanize suite passes (23 tests).


eeshsaxena:
Closing this to clear out my older open PRs, since it has been sitting a while without review. No hard feelings if it was not a priority. If it is still useful to you, say the word and I will reopen it and rebase it on current main.

PR #1312, opened 2026-07-06
fix quadratic backtracking in dehumanize unit search

arib06:
## Pull Request Checklist

## Description of Changes

drop the `\d` prefix alternative in the `dehumanize` per-unit search regex, since it let `re.search` re-enter a digit run at every offset so a long run of digits with no trailing unit backtracks the `\d+` against the missing literal and turns `dehumanize` O(n^2) on caller-supplied input (`'9'*50000` took ~14s; `^` and `\b` still anchor every real match, full suite green).


arib06:
any update?

PR #1316, opened 2026-07-09
restrict parser numeric fields to ASCII digits

arib06:
## Pull Request Checklist

## Description of Changes

The numeric fields in the parser (year, month, day, hour, minute, second, subsecond, ISO week and timezone offset) are matched with `\d`. In Python's default Unicode mode `\d` matches the whole Unicode decimal-digit category, not just ASCII `0-9`, so a value like `２０２１-０１-０２` or `٢٠٢١-٠١-٠٢` is accepted for a `YYYY-MM-DD` field even though those code points are not ISO 8601 digits. `int()` then happily converts them, so the non-ASCII string parses to a real date and the caller never sees that the input was not what the format claimed. I noticed it while comparing `arrow.get` against `datetime.fromisoformat`, which rejects the same strings, so untrusted date input that arrow validates can round-trip differently from the stdlib. Pinning these field regexes to `[0-9]` keeps the parser to ASCII digits and lines it up with `fromisoformat`. The timestamp tokens are left as-is and every existing format still parses (full suite green).


arib06:
any update?

PR #1325, opened 2026-07-14
reject out-of-range utc offset components in TzinfoParser

arib06:
## Pull Request Checklist

## Description of Changes

`TzinfoParser.parse` builds the offset from `int(hours) * 3600 + int(minutes) * 60` without checking either field's range, so `+0099` is silently accepted as `+01:39` (99 minutes folded into 1h39m) while `+2500`/`+9999` let a raw `datetime` `ValueError` escape instead of `ParserError`; rejecting `hours > 23` or `minutes > 59` in the parser keeps every valid offset working and turns the invalid ones into a normal parse error.


arib06:
any update?

PR #1329, opened 2026-07-21
Reject day of year 366 in a non-leap year instead of rolling over

Sreekant13:
## Pull Request Checklist

- [x] 🧪  Added **tests** for changed code.  (Updated the existing `test_YYYY_DDDD` assertion; see below.)
- [x] 🛠️  All tests **pass** when run locally.  (The parser test suite passes; I could not run the full `tox` matrix locally.)
- [ ] 🧹  All linting checks **pass** when run locally.  (Change is three source lines plus one test assertion; I did not run `tox -e lint` locally.)
- [ ] 📚  Updated **documentation** for changed code.  (N/A: no user-facing docs change.)

## Description of Changes

The DDD/DDDD tokens and `parse_iso` validate the day of year with `datetime.strptime("%Y-%j")`, which accepts day 366 in a non-leap year and rolls the result over into January 1 of the following year rather than rejecting it. That is inconsistent with day 367 and above (which are rejected), and surprising because a "day of year" input returns a date in a different year than the one parsed.

For example, `arrow.get("2023-366", "YYYY-DDD")` currently returns `2024-01-01`, while `arrow.get("2023-367", ...)` and `arrow.get("2023-456", ...)` are rejected with "The provided day of year N is invalid".

This detects the rollover and raises `ParserError`, matching the existing out-of-range handling: after `strptime`, if the resulting year differs from the parsed year, the day of year is out of range for that year. Leap years are unaffected, so `arrow.get("2024-366", "YYYY-DDD")` still returns `2024-12-31`.

Note this is a small behavior change. `test_YYYY_DDDD` previously asserted `parse_iso("2017-366") == datetime(2018, 1, 1)` with a comment describing the rollover as expected; that assertion is updated to expect a `ParserError`. The parser test suite passes.


Sreekant13:
Hi @jadchaar, gentle nudge on this whenever you have a moment. @Nishuuzz has since opened #1338 applying the same guard to ISO week 53, so this looks like a pattern rather than a one-off: the parser validates the range but never checks whether the result stayed in the requested year. Both are small behaviour changes, so happy to align this with whatever approach you prefer for the pair. Thanks for your work on arrow.

Sreekant13:
Hi @jadchaar, a gentle check-in on this one.

It rejects `arrow.get("2023-366", "YYYY-DDD")` instead of silently rolling it over to 2024-01-01, which is what `367` and `456` already do. The behaviour change is called out in the description, since it updates `test_YYYY_DDDD`, which had asserted the rollover as expected.

Two things that may help you judge it:

- The one red check, `windows-latest (pypy-3.11)`, fails on `master` as well. I checked the same job on the master branch rather than assume: ubuntu and macOS pypy pass, Windows pypy fails there too. So it is not this change.
- #1338 by @Nishuuzz applies the same guard to ISO week 53 in a 52-week year and cites this PR as the model, so there is a second instance of the pattern now.

No rush at all, and happy to rebase or split it if that helps.

PR #1334, opened 2026-08-08
Support decimal values in dehumanize

### LINKED ISSUE #1237: Make `dehumanize()` handle values with decimal fractions correctly
## Feature Request
Currently, `dehumanize()` is not capable of correctly parsing values with a floating point.
```py
>>> arrow.get("2025-12-10T09:00:00").dehumanize("2 days 3.5 hours ago")
<Arrow [2025-12-08T04:00:00+00:00]>
```
It seems that parser pretty much ignores the part up to the floating point.

A more expected result, however, would be that resulting Arrow object is `<Arrow [2025-12-08T05:30:00+00:00]>`

Not sure if there is a need to make floating point locale-dependent (ie, `.` or `,`), perhaps it'd be fine to allow both.

dchaudhari7177:
Fixes #1237.

## Problem

```python
>>> arrow.get("2025-12-10T09:00:00").dehumanize("2 days 3.5 hours ago")
<Arrow [2025-12-08T04:00:00+00:00]>   # 2 days 5 hours
```

The number pattern is `\d+`, so for `3.5 hours` the search lands on the digits *after* the separator and the `3` is dropped. Expected `<Arrow [2025-12-08T05:30:00+00:00]>`.

## Fix

The pattern becomes `\d+(?:[.,]\d+)?`, shared as `_NUMBER_PATTERN` between the two places that needed it (the per-timeframe search string and the number extraction). A matched value keeps parsing as `int` when it has no fraction, so nothing changes for the strings `humanize()` produces — only the new decimal case yields a `float`.

`relativedelta` accepts floats for seconds through weeks, so `shift()` needed no change.

## On the separator

Both `.` and `,` are accepted, as the issue suggests. The locale objects carry no decimal-separator information, and `dehumanize` input is written by hand rather than produced by `humanize()`, so there is nothing to key the choice off. The separator only counts when digits follow it, which is what keeps it from firing on a locale's own punctuation.

The trade-off is that digit grouping (`"1,500 hours"`) is read as `1.5`. That input is already wrong today — `\d+` matches just `1` — so this is not a regression, but it is not a case I tried to support; I noted it in a comment next to the pattern. Happy to restrict to `.` only if you would rather not accept the ambiguity.

## Fractional months and years

`relativedelta` rejects these outright (*"Non-integer years and months are ambiguous and not currently supported"*), which surfaces as a `ValueError` from `dehumanize`. That seemed better than silently truncating; there is a test pinning it.

## Tests

- `test_fractional_value` — hours, minutes and days, both directions.
- `test_fractional_value_comma_separator` — `1,5` matches `1.5`.
- `test_fractional_value_with_multiple_units` — the exact case from the issue, plus its future form.
- `test_fractional_months_and_years_are_rejected`.

Full suite passes (1903 passed, 1 skipped).

### DIFF
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -17,6 +17,7 @@ from time import struct_time
 from typing import (
     Any,
     ClassVar,
+    Dict,
     Final,
     Generator,
     Iterable,
@@ -1384,7 +1385,8 @@ class Arrow:
         current_time = self.fromdatetime(self._datetime)
 
         # Create an object containing the relative time info
-        time_object_info = dict.fromkeys(
+        # float as well as int: a value may carry a decimal fraction
+        time_object_info: Dict[str, Union[int, float]] = dict.fromkeys(
             ["seconds", "minutes", "hours", "days", "weeks", "months", "years"], 0
         )
 
@@ -1431,6 +1433,7 @@ class Arrow:
 
                 # If no number matches
                 # Need for absolute value as some locales have signs included in their objects
+                change_value: Union[int, float]
                 if not num_match:
                     change_value = (
                         1 if not time_delta.isnumeric() else abs(int(time_delta))
--- a/arrow/arrow.py
+++ b/arrow/arrow.py
@@ -74,6 +74,9 @@ _GRANULARITY = Literal[
     "year",
 ]
 
+# An unsigned integer or decimal, as accepted by dehumanize()
+_NUMBER_PATTERN = r"\d+(?:[.,]\d+)?"
+
 
 class Arrow:
     """An :class:`Arrow <arrow.arrow.Arrow>` object.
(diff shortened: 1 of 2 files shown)


Mukller (approved):
Verified locally against upstream master and this branch (Python 3.13, editable install):

**Master's decimal handling is silently wrong, this fixes it precisely**:

| input | master | branch |
|---|---|---|
| `"2 hours ago"` | −2h ✓ | −2h ✓ |
| `"1.5 hours ago"` | **−5h** (!) | **−1.5h** ✓ |
| `"0.5 days ago"` | **−5 days** (!) | **−0.5d** ✓ |
| `"1.5 days ago"` | — | −36h ✓ |

Master appears to split on the decimal point and treat fragments independently, so `"0.5"` becomes ~5 units and `"1.5"` becomes 1+5-ish depending on unit — a silent multi-hour/multi-day error for any user passing fractional strings. The `_NUMBER_PATTERN = r"\d+(?:[.,]\d+)?"` plus float-typed `time_object_info` fix handles both `.` and `,` separators correctly (verified `1.5` above; comma path covered by the added tests).

**Scope note**: I probed the future direction too — `"2 hours later"` raises `ValueError` identically on master *and* this branch, i.e. that's a pre-existing limitation of `dehumanize`'s grammar (`"in N units"` works on both), not something this PR regresses or needs to solve. Might make a good follow-up issue.

`pytest tests/test_arrow.py -k dehumanize` passes on the branch (the single collection error in my run was an unrelated fixture dependency).

Approving.

PR #1335, opened 2026-08-09
Fix IndexError in DateTimeParser.parse on malformed format strings

### LINKED ISSUE #1191: IndexError in DateTimeParser.parse

aldorizona10-glitch:
Fixes #1191.

### Problem
A malformed format string can leave a token in `fmt_tokens` without a corresponding capture group in the compiled pattern (the bracket/escape reinsertion in `_generate_pattern_re` can desync the two). `parse()` then called `match.group(token)`, which raised a raw `IndexError: no such group` instead of Arrow's documented `ParserMatchError`.

Minimal reproducer (the report's fuzzed PoC reduces to this):
```python
DateTimeParser().parse("foo", "[|(\\]s")   # IndexError: no such group
```

### Fix
Guard the group lookup in `parse()` so a missing group raises `ParserMatchError` — matching the handling already used a few lines below when a matched group's value is `None`. `parse()` already documents `:raises ParserMatchError:`, so this makes the behaviour consistent with the contract.

### Testing
Ran the parser suite locally: **724 passed**, plus a new regression test (`test_parse_malformed_fmt_no_match_group`). `black`, `flake8`, and `isort` are all clean.

---
*Implemented with AI assistance (Claude Code, Opus 4.8); I reviewed the change and verified it locally as described above. Happy to add a `CHANGELOG.rst` entry referencing this PR once it has a number.*


Mukller (approved):
Verified locally on the PR branch (Python 3.13, editable install):

**Before/after** for the malformed-format case from #1191:

```python
parser.DateTimeParser().parse("foo", "[|(\\]s")

release 1.4.0: IndexError: no such group          <- internal error leaks to users
branch:         ParserMatchError: Unable to find a match group for the specified token 's'.
```

`ParserMatchError` subclasses `ParserError`, so callers already handling arrow's documented parse failures catch this correctly — while previously a raw `IndexError` escaped the whole error hierarchy. Upstream issue #1191 is still open and this closes its crash surface; the regression test pins the exact repro.

`pytest tests/test_parser.py` — **722 passed, 1 skipped** (the coverage-threshold FAIL in my run is an artifact of running a single test file against tox.ini's 99% gate, not a test failure; full-suite CI covers it).

The try/except scope is tight — only the `match.group()` access is wrapped, so unrelated exceptions still propagate unchanged. Approving.

PR #1044, opened 2021-09-22
add Sinhala locale

ChingYi-AX:
## Pull Request Checklist

Thank you for taking the time to improve Arrow! Before submitting your pull request, please check all *appropriate* boxes:

- [x] 🧪  Added **tests** for changed code.
- [x] 🛠️  All tests **pass** when run locally (run `tox` or `make test` to find out!).
- [x] 🧹  All linting checks **pass** when run locally (run `tox -e lint` or `make lint` to find out!).
- [ ] 📚  Updated **documentation** for changed code.
- [x] ⏩  Code is **up-to-date** with the `master` branch.

If you have *any* questions about your code changes or any of the points above, please submit your questions along with the pull request and we will try our best to help!

## Description of Changes

Add Sinhala locale. Sinhala awares time frame format and describe function, takes into account the differences between general, past, and future forms (three different suffixes). So we do not copy the original timeframes dictionary as `timeframes_only_distance`, we wrote a new `timeframes_only_distance` dictionary :)

PR #1336, opened 2026-08-09
Fix misaligned Sinhala day abbreviations

Nishuuzz:
The weekday abbreviations in the Sinhala locale don't line up with the days they belong to. From Tuesday onwards they're off by a slot: Tuesday holds Wednesday's abbreviation, Thursday holds Friday's, Friday holds Saturday's, and Tuesday's own one has ended up down in the Saturday slot.

Two things go wrong because of it. Tuesday and Wednesday both come out as `බදා`, so you can't tell them apart, and Thursday's abbreviation is missing from the table altogether:

```python
>>> import arrow
>>> monday = arrow.Arrow(2026, 3, 9)
>>> [monday.shift(days=i).format("ddd", locale="si") for i in range(7)]
['සදුද', 'බදා', 'බදා', 'සිකු', 'සෙන', 'අ', 'ඉරිදා']
```

You don't have to read Sinhala to see that's wrong. Each abbreviation is meant to be the beginning of the day name it stands for, which is true for Monday, Wednesday and Sunday, but for Tuesday, Thursday, Friday and Saturday it's the beginning of a different day's name instead.

The fix is mostly just putting the existing entries back with the right days. Thursday was the one that had nothing to move back, so I've added `බ්‍රහ`, taken from the front of `බ්‍රහස්‍පතින්‍දා` the same way the others are formed from their day names.

I don't speak Sinhala, so I'd rather be clear about which bit is which. The reordering is mechanical and you can check it against the day names without knowing the language. The Thursday abbreviation is the one real judgement call, and if that isn't how it's normally shortened, tell me and I'll change it.

I also had to update `test_weekday`. It expected Saturday to abbreviate to `අ`, which is actually Tuesday's, so it was written to match the table rather than the language. That's presumably how this survived since the locale was added in #1044. The new `test_day_abbreviations` pins all seven, checks each one really is the start of its own day name, and checks they're all distinct.

Both tests fail before this change and pass after it. Full suite is green at 1902 passed, and black and flake8 are happy with both files.

One last thing while I was in here: Latvian uses `pi` for both `pirmdiena` and `piektdiena`, and Maltese uses `Ħ` for both `Il-Ħamis` and `Il-Ħadd`. Those are ambiguous rather than out of order, and choosing the right short forms needs someone who speaks the language, so I've left them alone. Happy to open an issue for them if it's worth tracking.


Mukller (approved):
Verified locally on the PR branch (Python 3.13): `pytest tests/test_locales.py -k "SinhalaLocale"` — **7/7 green**, including the new per-day abbreviation assertions.

The index-shift diagnosis is correct and the new mapping is linguistically consistent with the full day names:

| isoweekday | day_name | old abbr | new abbr |
|---|---|---|---|
| 1 Monday | සෙනසුරාදා-row | සදුද | සදුද |
| 2 Tuesday | | *(missing)* | අ |
| 3 Wednesday | | බදා | බදා |
| 4 Thursday | | බදා *(duplicate!)* | බ්‍රහ |
| 5 Friday | | සිකු | සිකු |
| 6 Saturday | | සෙන | සෙන |
| 7 Sunday | | ඉරිදා | ඉරිදා |

Old table had Tuesday and Thursday sharing "බදා" while Tuesday's own "අ" sat in the Sunday slot and Saturday's "සෙන" was duplicated into the weekend — i.e. everything from Tuesday onward was shifted by one. The new table aligns each abbreviation with its full name (අඟහරුවාදා→අ, බදාදා→බදා, බ්‍රහස්පතින්දා→බ්‍රහ), and the updated `test_weekday` assertion (Saturday → "සෙන", not "අ") pins the correction.

Note this intentionally changes output for Tue–Thu; any consumer hard-coding the old (wrong) abbreviations will see different strings — that's the point of the fix, but worth a changelog line as a behavior change.

PR #1339, opened 2026-08-20
Fix arrow.get() crash when tzinfo kwarg is explicitly None

mayuriphad:
## Summary

Fixes #1259.

`ArrowFactory.get()` decides whether kwargs should be routed to the direct `Arrow(...)` constructor (the 3+ positional args path) based on whether the only kwarg present is `tzinfo`. It did this by checking `tz is None`, which cannot distinguish between "`tzinfo` was not passed at all" and "`tzinfo=None` was passed explicitly". As a result:

```python
arrow.get('2013-01-01', 'YYYY-MM-DD', tzinfo=None)
```

was routed to `self.type(*args, **kwargs)` (the constructor path meant for calls like `arrow.get(2013, 5, 5, 12, 30, 45)`), which raised:

```
TypeError: Arrow.__init__() missing 1 required positional argument: 'day'
```

This is a real-world footgun for callers that pass an optional, possibly-`None`, timezone straight through as a kwarg (e.g. `arrow.get(value, fmt, tzinfo=account.timezone)` where `account.timezone` can be `None`).

## Fix

Check for the key's presence rather than its value:

```python
# tzinfo kwarg is not provided
    arg_count = 3
```

Now `tzinfo=None` behaves the same as omitting `tzinfo` entirely (defaults to UTC), while an explicit non-`tzinfo` single kwarg still correctly routes to the constructor path, and `tzinfo=<value>` still works as before.

## Test plan

- Added `test_kwarg_tzinfo_none_with_string_and_format` to `tests/test_factory.py`, asserting `arrow.get(str, fmt, tzinfo=None)` produces the same result as `arrow.get(str, fmt)` (UTC).
- `python -m pytest tests/test_factory.py -q` → 49 passed, 1 skipped.
- Manually verified before the fix this raised `TypeError`, and after the fix `arrow.get('2025-01-01', 'YYYY-MM-DD', tzinfo=None)`, `tzinfo='US/Pacific'`, and omitted-`tzinfo` all behave correctly.
- `python -m black --check arrow/factory.py tests/test_factory.py` → clean.


Mukller (approved):
Verified locally on the PR branch (Python 3.13, editable install) — real crash fixed, minimal correct change:

**Before/after** (release 1.4.0 vs branch):

```python
arrow.get("2013-01-01", "YYYY-MM-DD", tzinfo=None)

release: TypeError: Arrow.__init__() missing 1 required positional argument: 'day'
branch:  <Arrow [2013-01-01T00:00:00+00:00]>
```

Root cause confirmed by reading `factory.get`: the old guard `len(kwargs) == 1 and tz is None` cannot distinguish *absent* `tzinfo` from *explicitly passed* `tzinfo=None`, so the explicit-None call falls into the 3+-arg direct-constructor path with wrong arity. The new `"tzinfo" not in kwargs` condition is exactly the right distinction, and it preserves every other branch (verified: no-tz call returns an equal Arrow on the branch).

`pytest tests/test_factory.py` — green on the branch including the new regression test.

The single failing CI check is `windows-latest (pypy-3.11)` — unrelated to this one-line factory change; everything else is green.

Approving.

